PR #4, opened 2023-10-19
add postgresql setup

noworneverev:
### Short description

add postgresql setup to build the default role the database

### Proposed changes

add the ycms role and the database to the postgresql, otherwise the application cannot be started.
- 
- 

### Side effects

None
- 
- 

### Resolved issues

Fixes: cannot start the application with adding the database and role first


charludo:
Thanks, good point. Could you add this to the new WSL Readme? 

Also, I have zero clue about postgres tbh. None of this should be necessary with docker

Edit: also I think the package-lock.json got reformatted(?) try running prettier on it or remove the change from the commit

Issue #12, opened 2023-11-01
Tweak user groups and permissions for our usecase 


charludo:
### Motivation

Need to decide what user roles should exist, what permissions they should have, and what views they see by default after logging in, then reflect those decisions in the user model and login / registration flows.

### User Story

As a user I want to be able to log into a view that fits my needs.

### Proposed Solution

Use "Employee ID" and password

Possible roles:
Rollen: Stationsmanagement + Zentrales Bettmanagement + (Verwaltungsrolle)
- "Medical Personel": view patient data, enter new patient data (?)
- Stationmanagement: Assign rooms within their assigned stations, discharge patients (thus freeing up a bed)
- ZBM: See abstract occupancy information of all stations

### Alternatives

- Could also be an ID that an administrator-type user creates and can revoke for employees (not in the MVP though)
- Could make default view customizable for the users (also probably not in the MVP)

PR #31, opened 2023-11-02
Implement models

### LINKED ISSUE #11: Implement models for the MVP
### Motivation

Convert the proposed database scheme / ER diagram into Django models.

### Proposed Solution

Do it

### Alternatives

None

### Additional Context

https://drive.google.com/file/d/16Xa9yhgJekHPCLkNNJ13cDvq1qqZil5l/view

### Screenshots

image

charludo:
### Short description

Implemented all the models (except the user model, that happens in #12). Note that I had to make some changesto the ERD (see below for details), but the structure has largely remained the samer.

I' ve also taken the liberty to already implement some useful `cached_property`s in the models. (If you are unfamiliar with them: they are basically methods on the model objects that behave like attriutes, so you can use them just like you use the "real" attributes. For example, you can now use `ward.available_beds` to get the number of free beds in a ward, even though this is not stored in the database.) 

### Proposed changes

- implement the models
- adjust the test data fixture, tests, index view and patient form to match the new DB scheme 

### Side effects

- if you created a feature branch for your current issue, I highly recommend rebasing once this PR is merged, because there will be a lot of breaking changes

### Resolved issues

Fixes: #11 

image

Biggest differences to before: "Room" model to bridge the gap between wards/beds, and the User model is just how it currently is, will be changed to what we actually want in #12. Also a bunch of somewhat duplicate fields (like `is_private_room` on the `BedAssignment`) have instead been solved via a `cached_property`. 

And finally the `User` model should of course be connected 1:n to the `creator` field in each model, I've ommitted the connection so the picture would not get too cluttered, but of course it is done in the code 😄


noworneverev:
What's the difference between a ward and a room? Aren't they the same thing? In the diagram, one ward has multiple rooms. If so, the below screenshot demonstrate one of the wards, is it right?
image

charludo:
> What's the difference between a ward and a room? Aren't they the same thing? In the diagram, one ward has multiple rooms. If so, the below screenshot demonstrate one of the wards, is it right? 

Yes, exactly. That's one ward with 14 rooms. We need *some* way of differentiating between the two, and this just seemed the easiest. Without the `Room` model we could only state that there are 28 `Bed`s in this `Ward`, but not how they are organized. 

I was also thinking in terms of maybe later on (not the MVP, of course) allowing administrators to create a ward by setting its name etc, and the number of rooms; then afterwards just showing a list of those rooms, where they can set how many beds each room has.

noworneverev (approved):

PR #34, opened 2023-11-04
add flowbite library

noworneverev:
### Short description

Add flowbite library for further frontend development

### Proposed changes

Add flowbite library for further frontend development
- 
- 

### Side effects

Possbile overwritten of existing scss
- 
- 

### Resolved issues

Fixes: #

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "eslint-plugin-react": "^7.32.2",
         "eslint-plugin-react-hooks": "^4.6.0",
         "eslint-plugin-styled-components-a11y": "^1.0.0",
+        "flowbite": "^2.0.0",
         "jest": "^29.5.0",
         "lucide": "^0.215.0",
         "mini-css-extract-plugin": "^2.7.5",
@@ -73,18 +74,19 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.22.5",
+      "version": "7.22.13",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@babel/highlight": "^7.22.5"
+        "@babel/highlight": "^7.22.13",
+        "chalk": "^2.4.2"
       },
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.22.9",
+      "version": "7.23.2",
       "dev": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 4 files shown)


ycms/cms/templates/_base.html
```diff
@@ -18,72 +18,67 @@
...
+</div>
+{% endif %}
+<div class="py-6 mt-24 mx-12">
+    {% block content %}
+    {% endblock content %}
+</div>
 {% endblock raw_content %}
+{% block javascript %}
+<script src="https://cdnjs.cloudflare.com/ajax/libs/flowbite/2.0.0/flowbite.min.js"></script>
+{% endblock javascript %}
```
noworneverev: Oh, it works now, I think it's cached issue or something like that.

charludo: While this of course works, could you instead go with the method described here: https://github.com/themesberg/flowbite#init-functions ?

Basically, in `ycms/static/src/index.ts` add `import { initFlowbite } from 'flowbite'` at the very top, and `initFlowbite();` at the bottom inside the `DOMContentLoaded` eventlistener in that file?

That way we are not relying on a CDN, and Webpack should be able to bundle flowbite with our own JS for a smaller/faster package.

charludo: Could be!

I have added the example from the github to the index.html (temporarily), could you check if it works for you like this?

ycms/cms/templates/_base.html
```diff
@@ -18,72 +18,67 @@
...
             {% translate "Manage Patients" %}</a>
-            {% if user.is_authenticated %}
-                <div class="ml-auto flex-align-right font-bold hover:text-primary">
-                    <span class="text-primary">{{ user.first_name }} {{ user.last_name }}</span>
-                    <a href="{% url 'cms:public:logout' %}">({% translate "Logout" %})</a>
-                </div>
-                <div class="ml-4 mr-4 flex-align-right font-bold hover:text-white">
-                    <a href="{% url 'cms:protected:create_user' %}"
-                       class="text-primary hover:text-white">{% translate "Create User" %}</a>
-                </div>
```
charludo: Why is this gone? 😄

charludo (changes requested):
Thank you! See my comment below for two little things, but apart from that this looks good!

Two CircleCI flows are failing. Please run `./tools/code_style.sh` so djlint formats the `_base.html` file, and `./tools/translate.sh` to update the translations file.

ycms/cms/templates/_base.html
noworneverev: Yes, it works. I think that's it.

charludo: Awesome! I have pushed again to remove the example again, could you squash the commits? And then it's ready to merge!

charludo:
Same as before, the formatting in `_base.html` is broken, please use `./tools/djlint.sh`, and it would be great if you could squash your comments, see here: https://ycms.readthedocs.io/en/latest/git-workflow.html#clean-up-git-history before merging

charludo:
> Same as before, the formatting in `_base.html` is broken, please use `./tools/djlint.sh`, and it would be great if you could squash your comments, see here: https://ycms.readthedocs.io/en/latest/git-workflow.html#clean-up-git-history before merging

(Just change all `pick` in the beginning of the line to `fixup`, except for the first commit you did)

ycms/cms/templates/_base.html
noworneverev: The word is cut because my VSCode formatter, I'll fix it later.

ycms/cms/templates/_base.html
noworneverev: Don't work, not that simple. In its official Django document, it Include Flowbite’s JavaScript file inside the _base.html file just before the end of the <body> tag using CDN or by including it directly from the node_modules/ folder. https://flowbite.com/docs/getting-started/django/
I don't know how to include it from node_modules 😂

charludo (approved):
Awesome, thanks! 🎉

ycms/cms/templates/_base.html
charludo: Ah I see, it's a couple lines down, sorry for the confusion. But that appears to be the reason why the translation file has "changed".

ycms/cms/templates/_base.html
charludo: I have already tried this and it works 😅 

Is it OK with you if I add a commit to this branch and PR?

noworneverev: Really? I try but animantion only work partially, basically it only handle initialization when mounted. Yes you can commit.

charludo: Hmm, because of this: 
```
onMounted(() => {
    initFlowbite();
})
```
? I would just skip the `onMounted`, since we are not using vue anyways and the event handler should do essentially the same thing

charludo:
Ahh, sorry for the merge conflicts. If you want me to I' ll resolve them?

noworneverev:
> Ahh, sorry for the merge conflicts. If you want me to I' ll resolve them?

Ja, thank you!

charludo:
> > Ahh, sorry for the merge conflicts. If you want me to I' ll resolve them?
> 
> Ja, thank you!

ready to merge, the honor is yours :D

PR #40, opened 2023-11-07
Automatically infer bed availability

### LINKED ISSUE #23: Mark beds as unavailable

charludo:
### Short description

Currently, we require users to manually set a bed available/unavailable. This is not only errorprone and unconvenient, it is also unnecessary.

### Proposed changes

- replace the `is_available` attribute of `Bed`s with a cached property of the same name
- adjust the `available_beds` property of `Room`, because unfortunately, properties cannot be used in queryset filters.

### Side effects

- none, `is_available` can be used just as before, but is inferred automatically
- please check if my logic for determining bed availability is sound, though!!

### Resolved issues

Fixes: #23


noworneverev (approved):
no problem, looks nice.

PR #50, opened 2023-11-12
Add example bed assignments

### LINKED ISSUE #47: Add example bed assignemnts to test data fixture
### Motivation

Once #41 is merged, we will want our some pre-allocated assignments in our test data, so we don't have to constantly create them by hand. 

### Proposed Solution

Add a couple of assignments by hand to the test data.

charludo:
### Short description

Add bed assignments for the first 10 patients (ordered by admission date). Assignments were done by hand, but should reflect reality (private rooms for private insurance, no mixed gender rooms, roughly the same age for patients in each room).

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #47

### DIFF
--- a/ycms/cms/fixtures/test_data.json
+++ b/ycms/cms/fixtures/test_data.json
@@ -6105,7 +6105,7 @@
       "accompanied": false,
       "medical_record": 1,
       "recommended_ward": 1,
-      "bed": null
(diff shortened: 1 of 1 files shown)


ballerkind (approved):
👍

PR #51, opened 2023-11-12
Adjust base template and add sidebar

ballerkind:
### Short description

Adds a sidebar

### Proposed changes

-
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #30

PR #53, opened 2023-11-14
Role-based default views

### LINKED ISSUE #43: Set different default views for different roles
### Motivation

Users should always be redirected to the most appropriate view for them after logging in.
This includes:
- WardView afer logging in as a StationManager assigned to a specific station
- global ward editing view (if we have one) for ZBM
- and so on

charludo:
### Short description

Adds role/group based default views. After logging in, users will be redirected to their appropriate default view. Clicking on links for `cms:protected:index` will redirect in the same manner.

### Proposed changes

- move current index view to a `PatientsListView` and remove the outdated patient form from it
- refactor protected urls
- add group-based default views and a `UserBasedRedirectView`. This can also be used to instead redirect to a user's favorite view, should we implement the feature.

### Side effects

- none that I am aware of, there should not even be a conflict with #51 since the index route remains working

### Resolved issues

Fixes: #43


SnehaR26:
Hi, When I login as ZBM_000001 and click on Patient Intake tab, it gives 403 forbidden error. If this is our admin user, are we not allowing them to register new patients?

charludo:
> Hi, When I login as ZBM_000001 and click on Patient Intake tab, it gives 403 forbidden error. If this is our admin user, are we not allowing them to register new patients?

Yep, that's right. The idea is that ZBM are purely Administrators of the hospital, not involved with the patients, and should therefore not have access to patients' private data.

#51 will remove those links from the menu when you are logged in as ZBM

SnehaR26 (approved):
Looks good to me

PR #55, opened 2023-11-14
Feature/bed assignment management

### LINKED ISSUE #44: Patient assignment flow/view
### Motivation

In the WardView, a list of unassigned patients should be given and options for assigning them need to exist.

### Proposed Solution

- "Unassigned patients" box in the top row, showing number of unassigned and buttons for "manually assign" and "suggest assignment"
- "suggest assignment" should do nothing for now
- "manually assign" opens a sidepanel with unassigned users, with button "Assign" 
- clicking "Assign" closes the sidebar, colors room cards based on if the room is a good fir (age, gender, insurance status)

We probably sould show patient information somewhere while assigning them. It should also be possible to go againsty age and gender recommendations (with a warning)

noworneverev:
### Short description
Add Feature/bed assignment management

### Proposed changes

-  In ward view, users can edit the assignment and add a new assignment
-  Add another url [localhost/bed-assignments/](http://localhost:8086/bed-assignments/) to view and edit all bed assignments

### Side effects

- 
- 

### Resolved issues

Fixes: #44

### DIFF
--- a/ycms/cms/forms/bed_assignment/bed_assignment_form.py
+++ b/ycms/cms/forms/bed_assignment/bed_assignment_form.py
@@ -1,0 +1,51 @@
+import logging
+
+from django import forms
+from django.utils import timezone
+
+from ...models import BedAssignment
+from ..custom_model_form import CustomModelForm
+
+logger = logging.getLogger(__name__)
+
+
+class BedAssignmentForm(CustomModelForm):
+    """
+    Form for creating bed assignments.
+    """
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = BedAssignment
+        fields = ["medical_record", "admission_date", "discharge_date"]
+        widgets = {
+            "medical_record": forms.HiddenInput(),
+            "admission_date": forms.NumberInput(
+                attrs={"type": "date", "value": timezone.now().date}
+            ),
+            "discharge_date": forms.NumberInput(
+                attrs={
+                    "type": "date",
+                    "value": (timezone.now() + timezone.timedelta(days=1)).date,
+                }
+            ),
+        }
+
+    def save(self, commit=True):
+        """
+        This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm`
+        to create a new bed assignment.
+
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved bed assignment
+        :rtype: ~ycms.cms.models.bed_assignment.BedAssignment
+        """
+        if self.instance:
+            self.cleaned_data["medical_record"] = self.instance.medical_record
+        return super().save(commit)
--- a/ycms/cms/models/bed.py
+++ b/ycms/cms/models/bed.py
@@ -40,10 +40,16 @@ class Bed(AbstractBaseModel):
         :return: if the bed is available
         :rtype: bool
         """
-        return not self.assignments.filter(
-            models.Q(discharge_date__isnull=True)
-            | models.Q(discharge_date__gt=current_or_travelled_time())
-        ).exists()
+        if self.assignments.exists():
+            active_assignments = self.assignments.filter(
+                models.Q(admission_date__lte=current_or_travelled_time())
+                & (
+                    models.Q(discharge_date__gt=current_or_travelled_time())
+                    | models.Q(discharge_date__isnull=True)
+                )
+            )
+            return not active_assignments.exists()
+        return True
 
     def __str__(self):
         """
--- a/ycms/cms/models/room.py
+++ b/ycms/cms/models/room.py
@@ -74,8 +74,11 @@ class Room(AbstractBaseModel):
         patient_ids = BedAssignment.objects.filter(
             models.Q(bed__room=self)
             & (
-                models.Q(discharge_date__isnull=True)
-                | models.Q(discharge_date__gt=current_or_travelled_time())
+                models.Q(admission_date__lte=current_or_travelled_time())
+                & (
+                    models.Q(discharge_date__gt=current_or_travelled_time())
+                    | models.Q(discharge_date__isnull=True)
+                )
             )
         ).values_list("medical_record__patient", flat=True)
         patients = Patient.objects.filter(pk__in=patient_ids)
@@ -99,7 +102,57 @@ class Room(AbstractBaseModel):
         :return: genders of patients in the room
         :rtype: set
         """
-        return set(patient.gender for patient in self.patients.all())
+        return set(patient.gender for patient in self.patients())
+
+    @cached_property
+    def insurance_types(self):
+        """
+        Helper property for accessing all insurance_types of patients currently stationed in the room
+
+        :return: insurance_types of patients in the room
+        :rtype: set
+        """
+        return set(patient.insurance_type for patient in self.patients())
+
+    @cached_property
+    def patient_ages(self):
+        """
+        Helper property for accessing ages of patients currently stationed in the room
+
+        :return: ages of patients in the room
+        :rtype: list
+        """
+        return [patient.age for patient in self.patients()] if self.patients() else [0]
+
+    @cached_property
+    def minus_max_age(self):
+        """
+        Helper property for accessing minus maximum age of patients currently stationed in the room
+
+        :return: maximum age of patients in the room
+        :rtype: int
+        """
+        return -max(self.patient_ages)
+
+    @cached_property
+    def age_difference_between_patients(self):
+        """
+        Helper property for accessing age difference between patients currently stationed in the room
+
+        :return: age difference between patients in the room
+        :rtype: int
+        """
+        return max(self.patient_ages) - min(self.patient_ages)
+
+    @cached_property
+    def assignable_beds(self):
+        """
+        Helper property for accessing the free bed
+
+        :return: free beds in the room
+        :rtype: list
+        """
+        return [bed for bed in self.beds.all() if bed.is_available]
 
     def __str__(self):
         """
--- a/ycms/cms/templates/patient/patient_card.html
+++ b/ycms/cms/templates/patient/patient_card.html
@@ -1,0 +1,144 @@
+{% load i18n %}
+{% load widget_tweaks %}
+<div class="bg-white rounded-lg shadow-lg p-4 w-full">
+    <!-- Display Mode -->
+    {% if not standalone_mode %}
+        <div>
+            <!-- Patient Information Header -->
+            <div class="text-center mb-6">
+                <h2 class="text-2xl font-semibold">{% translate "Patient Information" %}</h2>
+            </div>
+            <!-- Patient Details -->
+            <div class="grid grid-cols-1 md:grid-cols-2 gap-4">
+                <!-- Personal Information -->
+                <div class="mb-2">
+                    <h3 class="text-lg font-semibold mb-2">{% translate "Personal Details" %}</h3>
+                    <p>
+                        <span class="font-semibold">{% translate "Name" %}:</span> {{ patient.first_name }} {{ patient.last_name }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Age" %}:</span> {{ patient.age }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Gender" %}:</span> {{ patient.gender | title }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Accompanied" %}:</span>
+                        {% if bed_assignment.accompanied %}
+                            <i icon-name="check" class="w-4 h-4 text-green-700"></i>
+                        {% else %}
+                            <i icon-name="x" class="w-4 h-4 text-red-700"></i>
+                        {% endif %}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Privately insured" %}:</span>
+                        {% if patient.insurance_type %}
+                            <i icon-name="check" class="w-4 h-4 text-green-700"></i>
+                        {% else %}
+                            <i icon-name="x" class="w-4 h-4 text-red-700"></i>
+                        {% endif %}
+                    </p>
+                </div>
+                <!-- Medical Information -->
+                <div class="mb-2">
+                    <h3 class="text-lg font-semibold mb-2">{% translate "Hostpital Stay Details" %}</h3>
+                    <p>
+                        <span class="font-semibold">{% translate "Diagnosis code" %}:</span>
+                        {{ bed_assignment.medical_record.diagnosis_code }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Note" %}:</span> {{ bed_assignment.medical_record.note }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Admission date" %}:</span>
+                        {{ patient.current_admission_date|default_if_none:'' }}
+                    </p>
+                    <p>
+                        <span class="font-semibold">{% translate "Discharge date" %}:</span>
+                        {{ patient.current_discharge_date|default_if_none:'' }}
+                    </p>
+                </div>
+            </div>
+            <!-- Buttons -->
+            {% if not assignment_mode %}
+                <div class="flex justify-end border-t pt-2 border-gray-200 rounded-b dark:border-gray-600">
+                    <div class="flex gap-2">
+                        <a href="{% url 'cms:protected:assign_patient' ward_id=ward.id assignment_id=bed_assignment.id %}">
+                            <button class="text-white bg-green-500 hover:bg-green-700 focus:ring-4 focus:outline-none focus:ring-green-300 dark:focus:ring-green-700  rounded inline-flex items-center px-5 py-2.5 text-center">
+                                {% if unassigned %}
+                                    {% translate "Assign" %}
+                                {% else %}
+                                    {% translate "Transfer" %}
+                                {% endif %}
+                            </button>
+                        </a>
+                        <button class="toggle-edit-button text-white bg-blue-500 hover:bg-blue-700 focus:ring-4 focus:outline-none focus:ring-blue-300 dark:focus:ring-blue-700  rounded inline-flex items-center px-5 py-2.5 text-center">
+                            {% translate "Edit" %}
+                        </button>
+                        {% if not unassigned %}
+                            <button id="dischargeButton-{{ ward.id }}"
+                                    data-modal-toggle="modalEl"
+                                    type="button"
+                                    data-bedassignment-id="{{ patient.current_stay.id }}"
+                                    class="focus:outline-none text-white bg-red-700 hover:bg-red-800 focus:ring-4 focus:ring-red-300 rounded px-5 py-2.5 dark:bg-red-600 dark:hover:bg-red-700 dark:focus:ring-red-900">
+                                {% translate "Discharge" %}
+                            </button>
+                        {% endif %}
+                    </div>
+                </div>
+            {% endif %}
+        </div>
+    {% endif %}
+    <!-- Edit Mode -->
+    <form enctype="multipart/form-data"
+          method="post"
+          action="{% url 'cms:protected:update_patient_stay' patient=patient.id|default:patient bed_assignment=bed_assignment.id|default:bed_assignment %}"
+          {% if not standalone_mode %}class="hidden"{% endif %}
+          data-unsaved-warning>
+        {% csrf_token %}
+        <!-- Patient Information Header -->
+        <div class="text-center mb-6">
+            <h2 class="text-2xl font-semibold">{% translate "Edit Patient Information" %}</h2>
+        </div>
+        <!-- Patient Details -->
+        <div class="grid grid-cols-1 md:grid-cols-2 gap-4">
+            <!-- Personal Information -->
+            <div class="mb-2">
+                <h3 class="text-lg font-semibold mb-2">{% translate "Edit Personal Details" %}</h3>
+                {% for field in patient_form %}
+                    <div>
+                        <label for="{{ field.id_for_label }}"
+                               {% if field.field.required %}class="field-required"{% endif %}>{{ field.label }}</label>
+                        {% render_field field %}
+                        <div class="help-text">{{ field.help_text }}</div>
+                    </div>
+                {% endfor %}
+            </div>
+            <!-- Medical Information -->
+            <div class="mb-2">
+                <h3 class="text-lg font-semibold mb-2">{% translate "Edit Hospital Stay Details" %}</h3>
+                {% for field in bed_assignment_form %}
+                    <div>
+                        <label for="{{ field.id_for_label }}"
+                               {% if field.field.required %}class="field-required"{% endif %}>{{ field.label }}</label>
+                        {% render_field field %}
+                        <div class="help-text">{{ field.help_text }}</div>
+                    </div>
+                {% endfor %}
+            </div>
+        </div>
+        <!-- Buttons -->
+        <div class="flex justify-end border-t pt-2 border-gray-200 rounded-b dark:border-gray-600">
+            <div class="flex gap-2">
+                <button type="submit"
+                        class="text-white bg-green-500 hover:bg-green-700 focus:ring-4 focus:outline-none focus:ring-green-300 dark:focus:ring-green-700  rounded inline-flex items-center px-5 py-2.5 text-center">
+                    {% translate "Save" %}
+                </button>
+                <button type="button"
+                        class="toggle-edit-button focus:outline-none text-white bg-red-700 hover:bg-red-800 focus:ring-4 focus:ring-red-300 rounded px-5 py-2.5 dark:bg-red-600 dark:hover:bg-red-700 dark:focus:ring-red-900">
+                    {% translate "Cancel" %}
+                </button>
+            </div>
+        </div>
+    </form>
+</div>
--- a/ycms/cms/templates/ward/ward_room_card.html
+++ b/ycms/cms/templates/ward/ward_room_card.html
@@ -1,23 +1,37 @@
 {% load i18n %}
-<div class="w-full max-w-sm p-6 border border-gray-200 rounded-lg shadow dark:bg-gray-800 dark:border-gray-700 {% if room.available_beds > 0 %}bg-green-300{% else %}bg-red-500{% endif %} {% if room.available_beds > 0 %}hover:bg-green-400{% else %}hover:bg-red-600{% endif %} hover:text-gray-900 cursor-pointer">
-    <h5 class="mb-2 text-2xl font-bold tracking-tight text-gray-900 dark:text-white">
-        {% translate "Room" %}
-        {{ room.room_number }}
-    </h5>
-    <ul class="text-base text-gray-800 list-disc list-inside dark:text-gray-400">
-        <li>{% translate "Beds" %}: {{ room.occupied_beds }} / {{ room.total_beds }}</li>
-    </ul>
-    {% if 'm' in room.genders and 'f' in room.genders %}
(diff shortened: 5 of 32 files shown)


charludo (changes requested):
Thank you! I have not had a closer look at the code yet, just tried it out locally, and everything seems to work!

However, I have some concerns about the workflow, especially the BedAssignmentForm.

After clicking on "Assign" in the sidebar, the form pops up and we can manually set a bed. But all the other fields are irrelevant to the situation, and it is not intuitive at all what bed we are assigning to, plus we don't know if there is someone of a different gender, or much older/younger, or privately insured in the room.

Could you rework this so no form is shown at all? The new workflow would be: 

- Click "assign" in the sidebar
- The sidebar closes, and the user can click on any of the rooms
- The patient is assigned to the first free bed in the room (e.g. use JS to submit the form in the background)
- The sidebar reopens

Additionally, we would need to indicate that we are currently in the "assignment mode":

- show something like "Assigning patient {{ patient name,... }}" somewhere
- in place of the "Unassigned Patients: x" + buttons show the patients relevant information (age, gender, diagnosis)
- visually indicate what rooms they are assignable to: set background and text color of private, other-gender, and large age difference rooms to grey, make full rooms or rooms with one person if the patient is accompanied unselectable. 
- same thing when editing an assignment, but also indicate the current room

I really like the indications (large age gap, insurance state and so on) but I think they, too, should only be shown during assignment - afterwards, they don't really matter.

charludo:
Also some other "nitpicks":

- Why not put the assignment information and action in the free space we already have, in the same style? 
image
- "New Bed Assignment" is not necessary, unless I am missing a usecase?
- I like the new sidebar! Could you also show the same diagnosis information in the room sidebar?
- Maybe we could also show insurance type in the assignment sidebar?
- Don't show discharge date in the assignment. In fact, I think we can also remove discharge date ffrom the room sidebar: Realistically, either you are already discharged (and no longer show up in the room sidebar), or you don't have a discharge date. I know we have the discharge date in the data, but AFAIK that is only because we get the complete dataset all at once, rather than linearly in time.

I know these are a LOT of change requests, I'm sorry 🙈

ycms/cms/forms/bed_assignment/bed_assignment_form.py
```diff
@@ -0,0 +1,106 @@
...
+                discharge_date=cleaned_data["discharge_date"],
+                accompanied=cleaned_data["accompanied"],
+                medical_record=cleaned_data["medical_record"],
+                recommended_ward=cleaned_data["recommended_ward"],
+                bed=cleaned_data["bed"],
+            )
+
+        if commit:
+            bed_assignment.save()
+        return bed_assignment
```
charludo: I don't think this does anything that the default save() method doesn't already do(?)

ycms/cms/views/bed_assignment/bed_assignment_create_view.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        # Check the referer to determine where the request is coming from
+        referer = self.request.META.get("HTTP_REFERER", "")
+        if "ward" in referer:
+            ward_id = referer.split("/ward/")[1].split("/")[0]
+            return reverse_lazy(
+                "cms:protected:ward_detail",
+                kwargs={"pk": 1 if not ward_id else int(ward_id)},
+            )
+        return reverse_lazy("cms:protected:manage_bed_assignment")
```
charludo: This + the UpdateView should just be this view, really. Inherit from TemplateView, and add `get()` and `post()` methods. If get, erturn rendered template+form. If post, check if the `/<pk:int>/`parameter is in the URL.

If it is not, create a new BedAssignment (though as mentioned above - I don't think we ever need to manually add a bed assignment, only implicitly at patient intake).

If it is, retrieve that object and assign it to `form.instance`.

Save form, return redirect to the bed assignments.

It's not that this solution doesn't work, it's just a lot of unnecessary/duplicate code.

charludo (commented):
Just some additional pointers :)

I also am not 100% convinced we even need the `BedAssignmentView`s - without being able to see the room occupations (see comments above), it is not a good idea to assign beds anyways.

So I'd argue we should only ever assign beds inside the `WardView`.

ycms/cms/models/timetravel_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+def set_request(**kwargs):
+    """
+    Helper function for setting the request
+    object once a signal is received
+    """
+    _thread_locals.request = kwargs["request"]
+
+
+request_signal.connect(set_request)
```
charludo: Quick question, are you merging the develop branch into your feature branch, or rebasing your feature branch onto the develop branch? I'm a bit confused why the develop changes always show up in your feature branch, it makes it a bit hard to see what was changed in your PR 😅 

Could you please rebase this PR onto develop so these changes disappear?

ycms/cms/forms/bed_assignment/bed_assignment_form.py
noworneverev: I've removed this. I think we can keep the `BedAssignmentViews` just for admin use. It's a bit inconvenient to view and edit bedassignment in the default admin page.

ycms/cms/views/bed_assignment/bed_assignment_create_view.py
noworneverev: I've removed the createview as it's not necessary.

noworneverev:
> Also some other "nitpicks":
> 
> * Why not put the assignment information and action in the free space we already have, in the same style?
>   image
> * "New Bed Assignment" is not necessary, unless I am missing a usecase?
> * I like the new sidebar! Could you also show the same diagnosis information in the room sidebar?
> * Maybe we could also show insurance type in the assignment sidebar?
> * Don't show discharge date in the assignment. In fact, I think we can also remove discharge date ffrom the room sidebar: Realistically, either you are already discharged (and no longer show up in the room sidebar), or you don't have a discharge date. I know we have the discharge date in the data, but AFAIK that is only because we get the complete dataset all at once, rather than linearly in time.
> 
> I know these are a LOT of change requests, I'm sorry 🙈

I redesign the room's sidepanel and limit that only admission date and discharge date can be modified at this point.

noworneverev:
> Thank you! I have not had a closer look at the code yet, just tried it out locally, and everything seems to work!
> 
> However, I have some concerns about the workflow, especially the BedAssignmentForm.
> 
> After clicking on "Assign" in the sidebar, the form pops up and we can manually set a bed. But all the other fields are irrelevant to the situation, and it is not intuitive at all what bed we are assigning to, plus we don't know if there is someone of a different gender, or much older/younger, or privately insured in the room.
> 
> Could you rework this so no form is shown at all? The new workflow would be:
> 
> * Click "assign" in the sidebar
> * The sidebar closes, and the user can click on any of the rooms
> * The patient is assigned to the first free bed in the room (e.g. use JS to submit the form in the background)
> * The sidebar reopens
> 
> Additionally, we would need to indicate that we are currently in the "assignment mode":
> 
> * show something like "Assigning patient {{ patient name,... }}" somewhere
> * in place of the "Unassigned Patients: x" + buttons show the patients relevant information (age, gender, diagnosis)
> * visually indicate what rooms they are assignable to: set background and text color of private, other-gender, and large age difference rooms to grey, make full rooms or rooms with one person if the patient is accompanied unselectable.
> * same thing when editing an assignment, but also indicate the current room
> 
> I really like the indications (large age gap, insurance state and so on) but I think they, too, should only be shown during assignment - afterwards, they don't really matter.

I have redesigned the assigning workflow.

ycms/cms/forms/bed_assignment/bed_assignment_form.py
```diff
@@ -0,0 +1,46 @@
+import logging
+
+from django import forms
+from django.utils import timezone
+from django.utils.translation import gettext_lazy as _
+
+from ...models import Bed, BedAssignment, MedicalRecord
+from ..custom_model_form import CustomModelForm
```
charludo:
```suggestion
from django import forms
from django.utils import timezone

from ...models import BedAssignment
from ..custom_model_form import CustomModelForm
```
Took me a while to figure out why pylint was passing despite these imports - you are missing an `__init__.py` in the `bed_assignment` directory!

ycms/cms/forms/bed_assignment/bed_assignment_form.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self, *args, is_update=False, **kwargs):
+        """
+        Initialize the form with additional filters.
+        1. Only show medical records that are not already assigned to a bed assignment.
+        2. Only show beds that are available.
+        """
+        super().__init__(*args, **kwargs)
+        self.is_update = is_update
+        self.fields["medical_record"].disabled = True
```
charludo:
```suggestion
```
I get what you are trying to do with the disabled field, but I think there's a cleaner way (see next comment)

ycms/cms/forms/bed_assignment/bed_assignment_form.py
```diff
@@ -0,0 +1,46 @@
...
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = BedAssignment
+        fields = ["medical_record", "admission_date", "discharge_date"]
+        widgets = {
+            "admission_date": forms.NumberInput(
```
charludo:
```suggestion
        widgets = {
            "medical_record": forms.HiddenInput(),
            "admission_date": forms.NumberInput(
```
This way the field will not be shown to the user at all.

ycms/cms/forms/bed_assignment/bed_assignment_form.py
```diff
@@ -0,0 +1,46 @@
...
+                attrs={"type": "date", "value": timezone.now().date}
+            ),
+            "discharge_date": forms.NumberInput(
+                attrs={
+                    "type": "date",
+                    "value": (timezone.now() + timezone.timedelta(days=1)).date,
+                }
+            ),
+        }
+
```
charludo, line 37:
```suggestion

    def save(self, commit=True):
        """
        This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm`
        to create a new bed assignment.

        :param commit: Whether or not the changes should be written to the database
        :type commit: bool

        :return: The saved bed assignment
        :rtype: ~ycms.cms.models.bed_assignment.BedAssignment
        """
        if self.instance:
            self.cleaned_data["medical_record"] = self.instance.medical_record
        return super().save(commit)
```
Disabled/Hidden fields can still be manipulated by the user; this ensures the medical_record of a bed assignment cannot be tampered with.

ycms/cms/models/bed.py
```diff
@@ -40,10 +40,19 @@ def is_available(self):
...
+            active_assignments = self.assignments.filter(
+                models.Q(admission_date__lte=timezone.now())
+                & (
+                    models.Q(discharge_date__gt=timezone.now())
+                    | models.Q(discharge_date__isnull=True)
+                )
+            )
+            return not active_assignments.exists()
+        # If no assignments, bed is available
+        return True
```
charludo: IMO the comments are pretty self-explanatory and we can remove them, but I'm also OK with leaving them. Your choice!

ycms/cms/models/room.py
```diff
@@ -74,8 +74,11 @@ def patients(self):
         patient_ids = BedAssignment.objects.filter(
             models.Q(bed__room=self)
             & (
-                models.Q(discharge_date__isnull=True)
-                | models.Q(discharge_date__gt=timezone.now())
+                models.Q(admission_date__lte=timezone.now())
+                & (
+                    models.Q(discharge_date__gt=timezone.now())
+                    | models.Q(discharge_date__isnull=True)
+                )
```
charludo: (since this is the same logic, but without comments)

ycms/cms/models/room.py
```diff
@@ -99,7 +102,61 @@ def genders(self):
...
+
+    @cached_property
+    def patient_ages(self):
+        """
+        Helper property for accessing ages of patients currently stationed in the room
+
+        :return: ages of patients in the room
+        :rtype: list
+        """
+        return [patient.age for patient in self.patients()]
```
charludo: How about this:
```suggestion
        return [patient.age for patient in self.patients()] if self.patients() else [0]
```
so we can then have an easier time with the other properties (see next comments)?

ycms/cms/models/room.py
```diff
@@ -99,7 +102,61 @@ def genders(self):
...
+    def minus_max_age(self):
+        """
+        Helper property for accessing minus maximum age of patients currently stationed in the room
+
+        :return: maximum age of patients in the room
+        :rtype: int
+        """
+        if patient_ages := self.patient_ages:
+            return -max(patient_ages)
+        return None
```
charludo: Then we can do:
```suggestion
        return -max(self.patient_ages)
```

ycms/cms/models/room.py
```diff
@@ -99,7 +102,61 @@ def genders(self):
...
+    def age_difference_between_patients(self):
+        """
+        Helper property for accessing age difference between patients currently stationed in the room
+
+        :return: age difference between patients in the room
+        :rtype: int
+        """
+        if patient_ages := self.patient_ages:
+            return max(patient_ages) - min(patient_ages)
+        return None
```
charludo: And also this:
```suggestion
        return max(self.patient_ages) - min(self.patient_ages)
```

ycms/cms/templatetags/custom_filters.py
```diff
@@ -7,3 +7,11 @@ def get_item(dictionary, key):
     Helper function for getting a dictionary item
     """
     return dictionary.get(key, "")
+
+
+@register.filter(name="times")
+def times(number):
```
charludo: would `range_` not be a more fitting name?
I am not quite sure why we need this anyways (see next comments) 😅

ycms/cms/templates/ward/ward_room_card.html
```diff
@@ -7,17 +8,39 @@ <h5 class="mb-2 text-2xl font-bold tracking-tight text-gray-900 dark:text-white"
...
-        <span class="bg-pink-100 text-pink-800 text-sm font-medium mr-2 px-2.5 py-0.5 rounded dark:bg-pink-900 dark:text-pink-300">
-        {% translate "Female" %}</span>
-    {% else %}
-    {% endif %}
+    <!-- For later fancy card use -->
+    <!-- <div class="relative">
+        {% for i in room.total_beds|times %}
+        <i icon-name="bed-single" class="w-16 h-16 text-white"></i>
+        {% endfor %}
+</div> -->
```
charludo:
```suggestion
    <!-- For later fancy card use -->
    <!-- <div class="relative">
        {% for _ in room.beds %}
        <i icon-name="bed-single" class="w-16 h-16 text-white"></i>
        {% endfor %}
</div> -->
```

But anyways, since this is not used in this/this PR, I would strongly advocate for removing this. Even if we need it later, currently it's just dead code.

ycms/cms/templates/bed_assignment/assign_room_card.html
```diff
@@ -0,0 +1,46 @@
+{% load i18n %}
+{% load custom_filters %}
```
charludo: As far as I can tell, this entire file is identical to `ward_room_card.html`. Why do we need it? 😅

ycms/cms/urls/protected.py
```diff
@@ -45,4 +50,31 @@
     ),
     path("ward/", ward.WardView.as_view(), name="ward_detail_default"),
     path("ward/<int:pk>/", ward.WardView.as_view(), name="ward_detail"),
+    path(
+        "discharge-patient/<int:assignment_id>/",
+        patients.DischargePatientView.as_view(),
+        name="discharge_patient",
+    ),
```
charludo:
```suggestion
```
Isn't this a duplicate of the `discharge_patient` route added in line 17?

ycms/cms/views/bed_assignment/bed_assignment_update_modal_view.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class BedAssignmentUpdateModalView(UpdateView):
+    """
+    View to update a patient in a modal
+    """
+
+    model = BedAssignment
+    template_name = "bed_assignment/update_bed_assignment_modal.html"
+    form_class = BedAssignmentForm
```
charludo: I am really unsure why we need this view, tbh. Correct me if I'm wrong, ut this is only used so that you can use JS to fetch the form for a specific bed assignment, correct?

Wouldn't it be a lot easier / less complicated to just include a form for each patient from the beginning?

Edit: actually, I have another easier idea. I will create a draft of that idea in a separate branch and ping you later. Ignore this for now 😄

ycms/cms/views/bed_assignment/bed_assignment_update_view.py
```diff
@@ -0,0 +1,49 @@
...
+    def get_form_kwargs(self):
+        """
+        Return the keyword arguments for instantiating the form
+
+        :return: The form kwargs
+        :rtype: dict
+        """
+        kwargs = super().get_form_kwargs()
+        kwargs["is_update"] = True
+        return kwargs
```
charludo:
```suggestion
```

ycms/cms/views/bed_assignment/bed_assignment_update_view.py
```diff
@@ -0,0 +1,49 @@
...
+        :rtype: str
+        """
+        # Check the referer to determine where the request is coming from
+        referer = self.request.META.get("HTTP_REFERER", "")
+        if "ward" in referer:
+            ward_id = referer.split("/ward/")[1].split("/")[0]
+            return reverse_lazy(
+                "cms:protected:ward_detail",
+                kwargs={"pk": 1 if not ward_id else int(ward_id)},
+            )
```
charludo: I cannot think of a better way to do this, but it still seems really hacky. The more I think about it, the more I am certain that we should drop the `/bed-assignments/` view alltogether. It complicates a lot of things (like this) without providing any benefit to the user.

Esp. since there is no way to create a patient or a medical record without a bed assignment, unless you are using the `/admin/`  backend, but you would still need to use the backend for patient+medical_record, at which point, why switch to this for the bed_assignment?

ycms/cms/views/patients/assign_view.py
```diff
@@ -0,0 +1,55 @@
...
+    context_object_name = "bed_assignment"
+
+    def get_context_data(self, **kwargs):
+        ward_id = self.kwargs["ward_id"]
+        assignment_id = self.kwargs["assignment_id"]
+
+        ward = Ward.objects.get(id=ward_id)
+        rooms = ward.rooms.all()
+        wards = Ward.objects.all()
+        # bed_assignment = BedAssignment.objects.get(id=assignment_id)
```
charludo:
```suggestion
```

ycms/cms/views/patients/assign_view.py
```diff
@@ -0,0 +1,55 @@
...
+            **super().get_context_data(**kwargs),
+        }
+
+    def post(self, request, *args, **kwargs):
+        """
+        This function assign a patient to a room
+        """
+        data = json.loads(request.body)
+        bed_id = data.get("bed_id")
+        assignment_id = kwargs["assignment_id"]
```
charludo, line 47: Haven't tested - but shouldn't this:
```suggestion
        bed_id = request.POST.get("bed_id")
        assignment_id = kwargs["assignment_id"]
```
also work?

ycms/static/src/index.ts
```diff
@@ -7,13 +7,23 @@ import "./js/utils/timetravel";
...
 
 window.addEventListener("DOMContentLoaded", () => {
     createIconsAt(document.documentElement);
     handleButtonDischarge();
     handleConfirmDischarge();
     handleModalClose();
+    handleRoomCardAssign();
+    handleAssignModalClose();
+    handleAssignModalConfirm();
+    handleButtonEdit();
```
charludo: Could we maybe handle these inside dedicated `load` eventlisteners inside the files? We would save ourselves having to export/import the functions and this eventlistener would not have to handle lots of unrelated tasks.

charludo (changes requested):
Some more comments 😅

ycms/cms/templates/patient/patient_card.html
```diff
@@ -0,0 +1,58 @@
...
+            </p>
+            <p>
+                <span class="font-semibold">{% translate "Admission date" %}:</span>
+                {% if patient.current_admission_date %}{{ patient.current_admission_date }}{% endif %}
+            </p>
+            <p>
+                <span class="font-semibold">{% translate "Discharge date" %}:</span>
+                {% if patient.current_discharge_date %}{{ patient.current_discharge_date }}{% endif %}
+            </p>
+        </div>
```
charludo: This is duplicate to `ward_drawer`.html. Please move into own partial temlate, then import in both of those.

ycms/cms/templates/patient/patient_card.html
```diff
@@ -0,0 +1,58 @@
...
+            <p>
+                <span class="font-semibold">{% translate "Diagnosis code" %}:</span>
+                {{ bed_assignment.medical_record.diagnosis_code }}
+            </p>
+            <p>
+                <span class="font-semibold">{% translate "Note" %}:</span> {{ bed_assignment.medical_record.note }}
+            </p>
+            <p>
+                <span class="font-semibold">{% translate "Admission date" %}:</span>
+                {% if patient.current_admission_date %}{{ patient.current_admission_date }}{% endif %}
```
charludo:
```suggestion
                {{ patient.current_admission_date }}
```
`None` values (or nonexistent attributes, for that matter) will simply not be rendered, ther is no advantage to checking like this.

There are a bunch of these, please change them all, I just picked this by chance 😄

charludo:
Some more things:

- could you also move the buttons into the unassigned patients card?
- I really likethe new cards inside the ward view sidebar! Could you also use them for the Assignment sidebar?
- could you solve the assignment flow without redirecting to a different (`/patients/assign/`) view? This view is 98% the same as the ward view. Could you instead just check with JS if we are currently assigning, and if so show the patient information at the top and color the rooms etc? (This probably can largely be done in CSS, not a lot of JS required). Then this view and template would not be needed anymore.
- speaking of: you currently show two cards at the top while assigning a ptient. The first one is superfluous IMO. Could you only show the second one, and make it the same height as the normal info cards, so the corridor and rooms don't shift when assigning a patient?

And lastly a general thing: It's obvious that you are putting a lot of work into this project, and that is awesome, thank you!! I think you could save yourself a large chunk of that work though if you started reusing the existing code more. A lot of the files you have included in this PR are almost duplicates of other files (I've mentioned a couple of them in the comments above). Slightly changing them would be way easier, and also keep our code way more maintainable.

ycms/cms/templatetags/custom_filters.py
noworneverev: I'll remove it.

ycms/cms/templates/ward/ward_room_card.html
noworneverev: Ok, I'll remove it.

ycms/cms/templates/bed_assignment/assign_room_card.html
noworneverev: The reason is that in ward view, the green card shows that there're still available beds, and red card simply indicate no beds available. In assigning page, the green one indicate no warning for that rooms, the red one indicate possible warnings(different gender, large age gap and mixed insurance), the grey one cannot be selected.
 I'll remove the duplicated code.

ycms/cms/views/patients/assign_view.py
noworneverev: No, not work.

ycms/static/src/index.ts
noworneverev: ok

ycms/cms/templates/patient/patient_card.html
noworneverev: It actually be rendered to string None which then shows on the page, it's not intuitive to see such word. 
image

charludo: That is very strange, same thing renders as empty string for me (and AFAIK it should generally not show None, unless somewhere we are returning the string "None").

I guess leave it for now, I'll try to investigate.

ycms/cms/templates/bed_assignment/assign_room_card.html
charludo: Ah, I see. That makes sense.

Although we could potentially even drop the colors in the normal/non-assign view completely, since staff probably don't need a constant warning "this room has 2 out of 2 patients in it", I think?

ycms/cms/templates/patient/patient_card.html
charludo: Ahhhhh, I get it now (I think, lol). `current_room`, `current_admission_date` and so on explicitly return None. Apparently, None is rendered as "None" (did not know that!).

On the other hand, things like `current_stay` will render as the empty string when the queryset is empty - I assume there's just some logic for that in the queryset's `__str__`.

I see three options:
- leave as is: works, but is a bit verbose IMO.
- use `{{ patient.current_admission_date|default_if_none="' }}` <- I would prefer this if you prefer to solve this in the template
- change the cached properties to return `""` instead of `None`; I don't think this should cause any troubles, since we are not testing for actual `None`-ness anywhere, only for falsy-ness. I would prefer this option, personally, as it is "set and forget", if we require those values somewhere else in a template it will just work

ycms/cms/templates/bed_assignment/assign_room_card.html
noworneverev: Agree, I'll remove the color.

charludo:
Two things I just noticed:
- clicking on edit in the ward sidebar right now creates an error 
- same as you mentioned in #65, some way to cancel the assignment/transfer workflows would probably be good. I know you can just go back in the browser, but users probably won't know.

ycms/cms/views/bed_assignment/bed_assignment_update_view.py
noworneverev: The view will be deleted after merge #65

noworneverev:
> Some more things:
> 
> * could you also move the buttons into the unassigned patients card?

done

> * I really likethe new cards inside the ward view sidebar! Could you also use them for the Assignment sidebar?

done

> * could you solve the assignment flow without redirecting to a different (`/patients/assign/`) view? This view is 98% the same as the ward view. Could you instead just check with JS if we are currently assigning, and if so show the patient information at the top and color the rooms etc? (This probably can largely be done in CSS, not a lot of JS required). Then this view and template would not be needed anymore.

Regarding the current solution to assign the patient in different view is that it involved some server side checkings to be able to display warnings for selected unassigned patient. It's not easy to fully depend on JS and CSS to deal with it. The current solution also opens possibility to assign the patient to a different ward. I would say after aligning the room cards, it's quite user friendly now, we can still open another issue to address redirecting url.

> * speaking of: you currently show two cards at the top while assigning a ptient. The first one is superfluous IMO. Could you only show the second one, and make it the same height as the normal info cards, so the corridor and rooms don't shift when assigning a patient?

done

> 
> And lastly a general thing: It's obvious that you are putting a lot of work into this project, and that is awesome, thank you!! I think you could save yourself a large chunk of that work though if you started reusing the existing code more. A lot of the files you have included in this PR are almost duplicates of other files (I've mentioned a couple of them in the comments above). Slightly changing them would be way easier, and also keep our code way more maintainable.

charludo:
> Regarding the current solution to assign the patient in different view is that it involved some server side checkings to be able to display warnings for selected unassigned patient. It's not easy to fully depend on JS and CSS to deal with it. The current solution also opens possibility to assign the patient to a different ward. I would say after aligning the room cards, it's quite user friendly now, we can still open another issue to address redirecting url.

Forgot to comment on this yesterday, but: You are absolutely right, I had not considered this!

charludo (approved):
🎉 🎉 🎉 🎉 🎉

PR #56, opened 2023-11-15
Add time travel functionality

### LINKED ISSUE #39: "Timetravelling" view
### Motivation

In th example data from moodle, registration-, admission-, and discharge dates are given for each patient, over a period of time.
In reality, patients come in one at a time, need to be assigned a bed, stay for an unknown amount of time, and are then, eventually, discharged, freeing up their bed for a new patient.

### Proposed Solution

I would keep our system working as intended for the real situation. However, we could add a subpath `/<date>/` for each url we have, showing the same views as the original view, but only loading and displaying data with a `created_date`, `admission_date`, `discharge_date`,... <= `<date>`.

In the nav bar/sidemenu we could then show the current simulated `<date>`, along with arrows to navigate forwards/backwards in time.

A `CustomManager` for the relevant models could help us with only retrieving relevant data without impeding the normal views.

### Alternatives

Alternatively, we could find some way of only loading in one piece of patient data at a time. This would still leave the problem of future discharge dates etc. though.

### Additional Context

I really wish Tabea gave us some more concrete info on this. Maybe whoever attends the next meeting can ask her about this?

charludo:
### Short description

This PR adds timetravel functionality: We can jump to any point in time, and see the patient data / hospital state at that time. This includes things like which patients exist, if they are currently staying at the hospital,...

### Proposed changes

- add `TimetravelManager` to provide queryset filtering while timetravelling
- add `TimetravelMiddleware` to notify manager are currently timetravelling
- add UI elements to control timetravelling
- adjust some `cached_properties` to work with this

### Side effects

- we need to take care how `cached_properties` are written in the future. When doing things like `.filter(patient__medical_records__bed_assignments__bed__room)`, django performs SQL joins under the hood - and these are not managed by the object manager, breaking our `TimetravelManager`'s functionality. Luckily, they can be rewritten to take advantage of the manager again.
- #51 should be merged first, so that I can adjust the links added in this PR
- probably some other stuff, this errs into "hacky" territory

### Resolved issues

Fixes: #39

### DIFF
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -120,7 +120,13 @@ limitations under the License.
                 </a>
             </li>
         </ul>
-        <div class="absolute bottom-0 w-full text-center text-gray-600 px-2">
+        <div class="absolute bottom-0 w-full text-center text-gray-600">
+            <li class="mb-4" id="timetravel-start">
+                <a href="?time=2024-01-01T00%3A00%3A00.000Z" class="justify-center"><i icon-name="timer-reset"></i> {% translate "Start Timetravel" %}</a>
+            </li>
+            <li class="hidden mb-4" id="timetravel-stop">
+                <a href="?" class="justify-center"><i icon-name="timer-off"></i> {% translate "Stop Timetravel" %}</a>
+            </li>
             <p class="text-sm">
                 {% translate "Logged in as" %} <span class="text-blue-500">{{ user.first_name }} {{ user.last_name }}</span>
             </p>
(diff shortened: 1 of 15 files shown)


ycms/cms/templates/_base.html
ballerkind, line 1: One small note: Could you remove the bullets of the li elements?

ballerkind (changes requested):
That is pretty impressive 👍

ycms/cms/templates/_base.html
charludo: Interesting - no bullets are shown for me in Firefox (see below), but indeed, there they are in Chromium 🙈 
image
Should be fixed now.

ballerkind (approved):

PR #62, opened 2023-11-19
Use custom time instead of timezone.now()

### LINKED ISSUE #39: "Timetravelling" view

charludo:
### Short description

#56 did not include a way of mocking `timezone.now()` when we were timetravelling. As a result, some functionality was broken, including in #55.

### Proposed changes

- add `current_or_travelled_time()` which returns the result of `timezone-now()` unless we are timetravelling, then it returns the travelled-to time.

### Side effects

- #55 will have to rebase on this

### Resolved issues

Fixes: #39


noworneverev (approved):
The `patients()` function in `room.py` currently contains a bug. It fails to return the correct patients if the same patient exists in bed assignments across multiple periods. I'll merge this to see if it resolves the issue; otherwise, we'll need to create a new issue to address this.

charludo:
> The `patients()` function in `room.py` currently contains a bug. It fails to return the correct patients if the same patient exists in bed assignments across multiple periods. I'll merge this to see if it resolves the issue; otherwise, we'll need to create a new issue to address this.

Interesting - does it just not rerurnany patient, or an actually wrong / unrelated patient? :O

noworneverev:
> > The `patients()` function in `room.py` currently contains a bug. It fails to return the correct patients if the same patient exists in bed assignments across multiple periods. I'll merge this to see if it resolves the issue; otherwise, we'll need to create a new issue to address this.
> 
> Interesting - does it just not rerurnany patient, or an actually wrong / unrelated patient? :O

For example, in bed assignment I have following records:
image

It should display the assignment admitted on Nov. 22, 2023 in ward page.
However, the room is correct, but patients inside that room got messed up.
image

PR #64, opened 2023-11-19
add responsive menu

### LINKED ISSUE #63: add responsive menu
### Motivation
Currently the menu is not responsive, it's not easy to use in small screen size.

### Proposed Solution
Add a responsive menu

noworneverev:
### Short description
Add responsive menu

### Proposed changes
Automatically minimize the the menu in small screen
- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #63

### DIFF
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -53,89 +53,96 @@ limitations under the License.
             {% endfor %}
         </div>
     {% endif %}
-    <div id="sidebar" class="absolute h-full w-64 bg-white border-r shadow-md">
-        <a href="{% url 'cms:protected:index' %}" class="block border-b">
-            <img src="{% static 'images/logo-ycms.png' %}"
-                 width="200px"
-                 height="auto"
-                 alt="{% translate " Logo" %}"
-                 class="my-3 mx-auto">
-        </a>
-        <ul>
-            <span class="block mt-5 px-5 text-gray-500 tracking-wide">{% translate "General" %}</span>
-            {% if perms.cms.add_ward %}
+    <div class="lg:flex">
+        <button id="sidebarToggle" class="block lg:hidden text-gray-600 px-4 py-3">
+            <i id="menuIcon" icon-name="menu"></i>
+            <i id="closeIcon" class="hidden" icon-name="x"></i>
+        </button>
+        <div id="sidebar"
+             class="fixed h-full w-64 bg-white border-r shadow-md transform -translate-x-full lg:transform-none lg:block">
+            <a href="{% url 'cms:protected:index' %}" class="block border-b">
+                <img src="{% static 'images/logo-ycms.png' %}"
+                     width="200px"
+                     height="auto"
+                     alt="{% translate " Logo" %}"
+                     class="my-3 mx-auto">
+            </a>
+            <ul>
+                <span class="block mt-5 px-5 text-gray-500 tracking-wide">{% translate "General" %}</span>
+                {% if perms.cms.add_ward %}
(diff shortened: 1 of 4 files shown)


ycms/static/src/js/utils/menu.ts
```diff
@@ -0,0 +1,8 @@
+window.addEventListener("load", () => {
+    const menuToggle = document.getElementById("menu-toggle");
+    const mobileMenu = document.getElementById("mobile-menu");
+
+    menuToggle?.addEventListener("click", () => {
+        mobileMenu?.classList.toggle("hidden");
```
charludo: `hidden` cannot be animated, sadly. How about `-translate-x-full`?

ycms/static/src/css/style.scss
```diff
@@ -278,3 +301,31 @@ input[type="submit"] {
...
+#mobile-menu li a.selected {
+    @apply pl-6 text-gray-800 bg-gray-200 border-blue-500;
+}
+
+#mobile-menu li a svg {
+    @apply w-[22px] mr-2;
+}
+
+#menu-toggle {
+    transition: all 0.3s ease;
```
charludo, line 314: I think you want to apply the transition duration to the mobile menu / sidebar itself, right?

ycms/cms/templates/_hamburger.html
```diff
@@ -0,0 +1,89 @@
+{% load i18n %}
+{% load static %}
+<!-- Hamburger menu for smaller screens -->
```
charludo: I appreciate the effort. However, there is really no need for a second menu. You can just apply the classes/styles/animations... to the existing sidebar, so we don't have duplicate structures.

ycms/cms/templates/_base.html
```diff
@@ -53,89 +53,93 @@
...
-                 height="auto"
-                 alt="{% translate " Logo" %}"
-                 class="my-3 mx-auto">
-        </a>
-        <ul>
-            <span class="block mt-5 px-5 text-gray-500 tracking-wide">{% translate "General" %}</span>
-            {% if perms.cms.add_ward %}
+    <div class="lg:flex">
+        <div id="sidebar"
+             class="fixed h-full w-64 bg-white border-r shadow-md hidden lg:block">
```
charludo:
```suggestion
             class="fixed h-full w-64 bg-white border-r shadow-md transform -translate-x-full lg:translate-x-0">
```
This way we can use the one sidebar on desktop and mobile, plus `translate-x` is animateable! :D

charludo (changes requested):
Thanks, this is a very good idea, and I'm impressed how good the site (esp. the ward view 😄) looks on mobile with this!

I have two thoughts:
- as mentioned in the comments below, there's no need for a second, mobile menu, since it just copies the structures from the desktop menu. Please just apply the classes/animations to the `sidebar`.
- Could you maybe also swap out the icon of the menu on click? Just would be a nice touch I think :D E.g. from icon name `menu` to `x` or something like that.

charludo (approved):
Very cool, thank you!

Only nitpick I have is it would be cool if the menu was still full height and the button would move to the right with it, but this absolutely works. SO if you have an idea for that feel free to implement it, otherwise go ahead and merge 😄

noworneverev:
> Very cool, thank you!
> 
> Only nitpick I have is it would be cool if the menu was still full height and the button would move to the right with it, but this absolutely works. SO if you have an idea for that feel free to implement it, otherwise go ahead and merge 😄

I think it's enough for now, we can improve it later.

PR #61, opened 2023-11-18
Adding a Ward Management View

SnehaR26:
### Short description

<br>
Adding the ward management view where combined data from all wards can be viewed and a new ward can be added

### Proposed changes

<br>
added a new view in ward/ward_management.py, a template to render it called ward_management.html, and made changes to protected.py and _base.html

- 

### Side effects

- 
- 

### Resolved issues

Fixes: #18

PR #65, opened 2023-11-23
Inline Discharge Update Form

charludo:
This is a suggestion to make the discharge date updating both more streamlined and userfriendly, and saving ourselves tons of semi-duplicate code.


noworneverev (commented):
Several cases need to be handled:
1. When the discharge date is modified to a date earlier than today, the entire card should disappear, and the ward page should refresh accordingly.
2. There's a lack of proper error handling for invalid dates, user might click the tick without selecting any date.
3. The inability to cancel edits needs to be addressed.
4. I think it is beneficial to allow modifications for other fields like admission date and accompanied status. We cannot assume they always fill it right in the first place.
5. Could you at least make /bed-assignments/ work? I still need that to overlook all bed assignments across multiple period which is beneficial for debugging purpose, it's too time consuming to select different time frame and I cannot get overall picture from that.

charludo:
> Several cases need to be handled

Yes, this is more a proof-of-concept, definitely not ready to be merged. Sorry, I should have made that clear in the PR description.

> 1. When the discharge date is modified to a date earlier than today, the entire card should disappear, and the ward page should refresh accordingly.

We could remove the "submit" functionality of the checkmark-button, and instead add an "apply" button at the top right which would submit (and then reload) all changed bed assignment forms - that way the user could go through multiple patients, changing dates, and then save all in one go?

> 2. There's a lack of proper error handling for invalid dates, user might click the tick without selecting any date.

True. Quick question, by default the current discharge date is selected though, right?

> 3. The inability to cancel edits needs to be addressed.

Would it be OK for you if clicking anywhere outside the checkmark button/input field would cancel the edit, or would you prefer a cancel button?

> 4. I think it is beneficial to allow modifications for other fields like admission date and accompanied status. We cannot assume they always fill it right in the first place.

Yes, agreed. The idea was that we could add forms of this kind to all relevant information. The JS is reusable, so it's just a matter of adapting the HTML form and adding a form handling view.

> 5. Could you at least make /bed-assignments/ work? I still need that to overlook all bed assignments across multiple period which is beneficial for debugging purpose, it's too time consuming to select different time frame and I cannot get overall picture from that.

OK, I had not thought about it in this way. That makes sense. My main issue with the way it is currently handled (or was handled, I haven't reviewed #61 again yet) are the almost-duplicate routes/forms/views.


Before I put more time into this: Would you like to go in this direction with the inline forms vs the modal? (After addressing all your concerns above, of course?)

noworneverev:
> > Several cases need to be handled
> 
> Yes, this is more a proof-of-concept, definitely not ready to be merged. Sorry, I should have made that clear in the PR description.
> 
> > 1. When the discharge date is modified to a date earlier than today, the entire card should disappear, and the ward page should refresh accordingly.
> 
> We could remove the "submit" functionality of the checkmark-button, and instead add an "apply" button at the top right which would submit (and then reload) all changed bed assignment forms - that way the user could go through multiple patients, changing dates, and then save all in one go?
> 

That would be too complex, and it's not a common case to modify many patients at once. I think we can preserve the edit button, after clicking the edit button, `Transfer` and `Discharge` buttons are hided, and show buttons `Save` and `Cancel`, then some fields are opened for modification, user can finally decided whether to save or cancel the edition.

> > 2. There's a lack of proper error handling for invalid dates, user might click the tick without selecting any date.
> 
> True. Quick question, by default the current discharge date is selected though, right?
> 

No default date are selected after I click the pencil icon.
image


> > 3. The inability to cancel edits needs to be addressed.
> 
> Would it be OK for you if clicking anywhere outside the checkmark button/input field would cancel the edit, or would you prefer a cancel button?
> 

I prefer a cancel button, or after user leaves the drawer of selected room, the edit mode should be cancelled.

> > 4. I think it is beneficial to allow modifications for other fields like admission date and accompanied status. We cannot assume they always fill it right in the first place.
> 
> Yes, agreed. The idea was that we could add forms of this kind to all relevant information. The JS is reusable, so it's just a matter of adapting the HTML form and adding a form handling view.
> 
> > 5. Could you at least make /bed-assignments/ work? I still need that to overlook all bed assignments across multiple period which is beneficial for debugging purpose, it's too time consuming to select different time frame and I cannot get overall picture from that.
> 
> OK, I had not thought about it in this way. That makes sense. My main issue with the way it is currently handled (or was handled, I haven't reviewed #61 again yet) are the almost-duplicate routes/forms/views.
> 
> Before I put more time into this: Would you like to go in this direction with the inline forms vs the modal? (After addressing all your concerns above, of course?)

Yeah, this direction is good, it's more user friendly.

charludo:
> I think we can preserve the edit button, after clicking the edit button, `Transfer` and `Discharge` buttons are hided, and show buttons `Save` and `Cancel`, then some fields are opened for modification, user can finally decided whether to save or cancel the edition.

That's a great idea, I'll implement this and let you know!

> No default date are selected after I click the pencil icon.

Hm, very strange, for me it's working. I'll have another look.

charludo:
OK, this is ready for review @noworneverev @ballerkind @SnehaR26.

- allow editing of all patient & bed assignment info in an inline-form, reusing the existing Form classes
- use this form for the `/bed-assignments/.../.../` as well (known bug: saving a patient/assignment via this list will not redirect to the bed assignments list; IMO not worth it to find a workaround, since this is only for internal use)
- added an unassigned patient to the test data fixture for easier testing
- fix required fields star offset (unrelated to this PR, it was just bothering me, lol)
- multiple files are deleted and their functionality handled in another / an existing view

This PR is hard to review. I would suggest maybe just trying every function out, and having a lquick look at the code for any glaring mistakes? Since this does not merge into `develop`, but into the feature branch for #55.

PR #66, opened 2023-11-28
Add extended test data

charludo:
### Short description

Add more test data

### Proposed changes

- add 5 new patients and 7 new medicalrecords and bed assignments, incl. one patient with 3 separate stays in past/present/future
- these are added in a separate fixture file, just so we can easily disable them for e.g. the presentation
- add new 1 ward (Pädiatrie) with 4 rooms and 8 beds, where one room only has 1 bed and one has 3
- added new bed type "small" used for the children's beds 

### Side effects

- the issues @noworneverev mentioned regarding displayed vs actual bed assignments are very obvious in this. I think #55  will already solve some of that though, right? **Edit:** yes, together with #67
- I made no attempt at bug hunting in this PR, I think that is going to be a separate story...

### Resolved issues

Fixes: n/a


noworneverev (approved):
looks good

PR #67, opened 2023-11-28
Fix the current_stay property (again)

charludo:
### Short description

Fix the return value of `patient.current_stay`

### Proposed changes

- implement stay filtering logic analogous to the ones in the `Bed`, `Room` and `Ward` models implemented in #55  for the `Patient`'s `current_stay`

### Side effects

- this, on its own, won't completely solve the issue when just looking at patients in the ward view, merging #55 and this should though - at least as far as I have been able to test!

### Resolved issues

Fixes: n/a


noworneverev (approved):
Looks good, I'll rebase this to #55 to see if it solve the issue.

PR #68, opened 2023-11-29
Update patient management view

### LINKED ISSUE #52: Rework patient management view
### Motivation

The view is currently just the example index view (outdated form for patient creation and list of patients, with missing attributes). This should become a useful view e.g. for medical personnel wanting to add medical records, or for us wanting to add a new patient without a medical record and bed assignment.

### Proposed Solution

Add a (ideally filterable) CRUD table with action buttons for delete, edit (maybe inline?), add medical record

ballerkind:
### Short description

Updates the patient management view

### Proposed changes

- view all patients in a table
- inline editing

### Side effects

- 
- 

### Resolved issues

Fixes: #52

### DIFF
--- a/ycms/cms/templates/patients_list.html
+++ b/ycms/cms/templates/patients_list.html
@@ -7,34 +7,142 @@
     {% translate "Manage Patients" %}
 {% endblock title %}
 {% block content %}
-    <div class="max-w-6xl mx-auto">
-        {% if perms.cms.view_patient %}
-            <div class="w-full flex flex-col flex-wrap mb-6">
-                <h1 class="heading">{% translate "Patients list" %}</h1>
-            </div>
-            {% for patient in patients %}
-                <div class="bg-white border-2 justify-between border-gray-200 rounded-lg shadow-md shadow-inner mb-8 gap-6 overflow-hidden">
-                    <div class="px-8 py-6 grow">
-                        <p class="text-xl font-bold">
-                            <span class="font-bold">{{ patient.last_name }}</span>, {{ patient.first_name }}
-                        </p>
-                        <p>
-                            {% translate "Privately insured" %}:
-                            {% if patient.insurance_type %}
-                                <i icon-name="check" class="w-4 h-4 text-green-700"></i>
-                            {% else %}
-                                <i icon-name="x" class="w-4 h-4 text-red-700"></i>
-                            {% endif %}
-                        </p>
-                        <p class="text-gray-800 mt-4">{% translate "DOB" %}: {{ patient.date_of_birth }}</p>
-                        <p class="text-gray-800">{% translate "Age" %}: {{ patient.age }}</p>
-                        <p class="text-gray-800 mb-4">{% translate "Gender" %}: {{ patient.gender }}</p>
-                        <p class="text-gray-600">{% translate "Added by" %}: {{ patient.creator.personnel_id }}</p>
-                        <p class="text-gray-600">{% translate "Added at" %}: {{ patient.created_at }}</p>
-                        {{ patient.current_stay }}
-                    </div>
-                </div>
-            {% endfor %}
-        {% endif %}
-    </div>
+    {% if perms.cms.view_patient %}
+        <div class="w-full flex flex-col flex-wrap mb-6">
+            <h1 class="heading">{% translate "Patients list" %}</h1>
+        </div>
+        <div class="shadow-md">
+            <input id="search-patient-input"
+                   type="text"
+                   placeholder="Search for name..." />
+            <table id="patients" class="w-full text-sm text-gray-500">
+                <thead class="text-xs text-gray-700 uppercase bg-gray-100">
+                    <tr>
+                        <th class="px-6 py-3">{% translate "Patient name" %}</th>
+                        <th class="px-6 py-3">{% translate "Gender" %}</th>
+                        <th class="px-6 py-3">{% translate "Date of Birth" %}</th>
+                        <th class="px-6 py-3">{% translate "Privately insured" %}?</th>
+                        <th class="px-6 py-3 w-60">{% translate "Status" %}</th>
+                        <th class="px-6 py-3 w-[400px]">{% translate "Action" %}</th>
+                    </tr>
+                </thead>
+                <tbody class="overflow-y-auto">
+                    <tr class="odd:bg-white even:bg-gray-100 border-b">
+                        <td class="px-6 py-4">
+                            <div class="flex flex-nowrap gap-4">
+                                {% render_field new_patient_form.last_name form="add-patient" %}
+                                {% render_field new_patient_form.first_name form="add-patient" %}
+                            </div>
+                        </td>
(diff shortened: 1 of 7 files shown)


ycms/cms/templates/patients_list.html
```diff
@@ -7,34 +7,181 @@
...
+                   onkeyup="search()"
+                   type="text"
+                   placeholder="Search for name..." />
+            <table id="patients" class="w-full text-sm text-gray-500">
+                <thead class="text-xs text-gray-700 uppercase bg-gray-100">
+                    <tr>
+                        <th class="px-6 py-3">{% translate "Patient name" %}</th>
+                        <th class="px-6 py-3">{% translate "Gender" %}</th>
+                        <th class="px-6 py-3">{% translate "Age" %}</th>
+                        <th class="px-6 py-3">{% translate "Private" %}?</th>
```
charludo:
```suggestion
                        <th class="px-6 py-3">{% translate "Private Insurance" %}?</th>
```

ycms/cms/templates/patients_list.html
```diff
@@ -7,34 +7,181 @@
...
+                        <td class="px-6 py-4 text-center">
+                            <input name="insurance_type"
+                                   form="add-patient"
+                                   type="checkbox"
+                                   {% if patient.insurance_type %}checked{% endif %} />
+                        </td>
+                        <td class="px-6 py-4"></td>
+                        <td>
+                            <button class="btn" form="add-patient">{% translate "Add" %}</button>
+                        </td>
```
charludo: OK, this works very well - but is there a reason you couldn't just reuse the `PatientForm`? That would simplify things significantly, both here and in the view, and also provide serverside validation.

For editing this is slightly trickier, but in theory you could either keep the custom HTML form with the same naming conventions as the `PatientForm` and just use the `PatientForm` for backend-validation/model saving, or you could pass a form for each patient (or a Patient modelformset).

ycms/cms/views/patients/patients_list_view.py
```diff
@@ -27,4 +30,79 @@ def get_context_data(self, **kwargs):
...
+
+        else:
+            patient = Patient.objects.create(
+                creator=request.user,
+                first_name=first_name,
+                last_name=last_name,
+                insurance_type=insurance_type,
+                gender=gender,
+                date_of_birth=date_of_birth,
+            )
```
charludo: This is what I meant in the comment above, this should just be handled by the `PatientForm`.

ycms/cms/views/patients/patients_list_view.py
```diff
@@ -27,4 +30,79 @@ def get_context_data(self, **kwargs):
...
+            patient = Patient.objects.get(id=request.POST["id"])
+
+            messages.success(
+                request,
+                _('Patient "{} {}" has been deleted.').format(
+                    patient.first_name, patient.last_name
+                ),
+            )
+
+            patient.delete()
```
charludo:
```suggestion
            patient.delete()
            
            messages.success(
                request,
                _('Patient "{} {}" has been deleted.').format(
                    patient.first_name, patient.last_name
                ),
            )
```
veeery nitpicky, but technically... :D

ycms/cms/views/patients/patients_list_view.py
```diff
@@ -27,4 +30,79 @@ def get_context_data(self, **kwargs):
...
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: Redirect to list of patients
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+
+        if len(request.POST) == 2:
+            patient = Patient.objects.get(id=request.POST["id"])
```
charludo: This is a bit confusing tbh. Could you maybe add another input to the form with the name "delete" (or something like that) and check for its presence? Just from these two lines, it's not at all obvious why you would check the length. And if we ever add another form with just one field to this page, the bug hunt would not be fun.

ycms/cms/templates/patients_list.html
```diff
@@ -7,34 +7,181 @@
...
+                                    {% translate "Save" %}
+                                </button>
+                            </td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+            <script>
+                // TODO: move to typescript file
+                function editPatient(id) {
```
charludo:
```suggestion
                const editPatient = (id: string) => {
```
same below, eslint will complain otherwise once you move this to its proper file

ycms/cms/templates/patients_list.html
```diff
@@ -7,34 +7,181 @@
...
+                            <td>
+                                <button class="btn" onclick="editPatient({{ patient.id }})">{% translate "Edit" %}</button>
+                                <button class="btn hidden" onclick="editPatient({{ patient.id }})">{% translate "Cancel" %}</button>
+                                <button class="btn hidden" form="delete-patient-{{ patient.id }}">
+                                    <form hidden id="delete-patient-{{ patient.id }}" method="post">
+                                        {% csrf_token %}
+                                        <input type="text" name="id" value="{{ patient.id }}" />
+                                    </form>
+                                    {% translate "Delete" %}
+                                </button>
```
charludo: Had no idea buttons could contain forms. Nice.

IMO Delete should be moved to "top level" of buttons though, given red color and maybe a confirmation popup (see ward view for comparison)?

ycms/cms/templates/patients_list.html
```diff
@@ -7,34 +7,181 @@
...
+                            <td class="px-6 py-4">
+                                {% if patient.gender == 'm' %}
+                                    <p class="py-2 px-3 pr-10 text-xl border border-transparent w-full bg-blue-100 text-blue-800 text-sm font-medium px-2.5 py-0.5 rounded dark:bg-blue-900 dark:text-blue-300">
+                                        {% translate "Male" %}
+                                    </p>
+                                {% else %}
+                                    <p class="py-2 px-3 text-xl border border-transparent w-full bg-pink-100 text-pink-800 text-sm font-medium px-2.5 py-0.5 rounded dark:bg-pink-900 dark:text-pink-300">
+                                        {% translate "Female" %}
+                                    </p>
+                                {% endif %}
```
charludo: diverse as third option would be nice! Since its also in the models/constants.

charludo (changes requested):
Thanks!! 🎉 

Very nice touch with the status linking to the ward.

I have a couple small-ish comments, and also two suggestions you could implement if you want - otherwise, a backlog issue for each would be great!

- sorting by clicking the table headers
- after editing/adding a patient, highlight the resulting row after the redirect (e.g. for a couple of seconds with a css animation), and also scroll to the just-edited entry.

ycms/cms/views/patients/patients_list_view.py
charludo: Edit: or even just add another method that handles DELETE requests. IMO even cleaner.

charludo:
Alright, as far as I'm concerned this is ready to merge. Please rebase on develop (`resolve_translation_conflicts.sh` can be used during rebase to solve the current merge conflict) and squash

charludo (approved):
🎉

PR #76, opened 2023-11-30
Fix: BedAssignment Date fields should be DateTime

### LINKED ISSUE #75: Change Admission- and Discharge from Date fields to DateTime fields
### Describe the Bug

Just noticed that in the original data, admission/discharge fields contain a time. Since we are using `DateField`s in the `BedAssignment` model, that information got lost

### Additional Information

Solution: recreate the testdata fixture...... 
<details>
  <summary>Traceback</summary>

  ```
  ```
</details>

charludo:
### Short description

Re-added the time information that was lost when originally importing the patient data into our DB.

### Proposed changes

- 
- 

### Side effects

- I hope these are all correct. If you could check one or two random entries from the data.csv on moodle and this one, that would be fantastic.
- 

### Resolved issues

Fixes: #75


SnehaR26:
So checked for 3 patients and the data in our Db matches with Moodle data for admission/discharge date and time.

SnehaR26 (approved):

PR #83, opened 2023-12-04
Fix: patients in unassigned sidebar should also be editable

charludo:
### Short description

I missed that the new inline form was not showing in the unassigned sidebar. This PR fixes that.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: n/a


SnehaR26 (approved):
Inline form is visible in unassigned drawer and works as intended. 😊

PR #89, opened 2023-12-05
Fix sidebar visibility

### LINKED ISSUE #72: Sidebar visible while resetting / first setting the password
### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```
</details>

ballerkind:
### Short description

Remove sidebar where not needed

### Proposed changes

- introduce new file "_interface.html" and new block "interface_content"
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #72

### DIFF
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -18,41 +18,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 {% endcomment %}
 {% block raw_content %}
-    {% if messages and not messages_suppressed %}
-        <div id="messages"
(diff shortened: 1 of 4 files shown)


charludo (changes requested):
Alright, this certainly works! 😄 

Maybe we could keep the diff smaller though, if you move the messages (so basically everything left inside `_base.html`) to `_raw.html`, and just let the "no-sidebar" views inherit from there instead?

Just a thought though, please do correct me If you have tried this and there's a complication I'm not seeing!

charludo (approved):

PR #54, opened 2023-11-14
Add tests for the WardView

charludo:
### Short description

Test `WardView` functionality

### Proposed changes

- add tests
- fix permission issue

### Side effects

- we are at 80% test coverage! 😄🎉 

### Resolved issues

Fixes: #48

PR #91, opened 2023-12-06
Fix tests failing on new data

charludo:
### Short description

We introduced new test data since #54 was originally opened. So now some tests needed to be adjusted.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #


SnehaR26 (approved):
Code changes look ok (some review required but that's for later PR).

PR #93, opened 2023-12-06
Add timeline view

### LINKED ISSUE #59: Timeline view for the ward
### Proposed Solution

- add timeline view mode in the WardView
- allow dragging patients between rooms, but not between times
- use visjs timelines

charludo:
### Short description

TIMELINE 🎉 

### Proposed changes

- add a user setting which saves whether they should see the ward as ward view or timeline mode. This is the most consistent user experience I could come up with, and also one where a lot of parts from the ward view templates can be reused.
- add a timeline view. This displays all bed assignments on a timeline, and allows users to drag&drop patients on the timeline.
- bulk save functionality: I thought about using ajax to save the assignments after each drag&drop without reloading the page. However, there were a couple of disadvantages with that, and one big advantage with the bulk method: When we build the " suggest assignment" feature, alomst all the work is already done. All we need to do is change the "source" data to the one suggested by the algorithm, and saving the suggestion will already work :) 

### How to review this

- please try out the timeline with any and all edgecases you can think of.
- there's two pieces of code which need special attention:
    - the timeline js framework does not limit the number of people per room. That is custom logic, and it's *hairy*. Please check the logic of the ` onMoving()`  and `maxConcurrentBeds()`  functions in `timeline.ts`. It is likely that an error has occurred somewhere in there.
    - the same goes for the assignment logic in the `post()` method of the `TimelineView`. 

### Resolved issues

Fixes: #59, partially: #119


ycms/cms/templates/_raw.html
```diff
@@ -67,6 +67,7 @@
                         </div>
                     {% endif %}
                     {% if msg.level_tag == 'error' %}
+                        <<<<<<< HEAD
```
ballerkind: Git leftover

ballerkind (changes requested):
First of all, this is a very nice feature that works pretty well 👍

While testing the timeline, I noticed some things:
- It would be nice to have a "Cancel" button, maybe next to the "Save" button

- After moving multiple patients between different rooms without saving in between, I lost the overview of what I have changed in total. Maybe we can mark the edited but not yet saved assignments somehow (different background or border color?)

- The sidebar entry "Ward View" is not marked as "selected" when we are at the timeline view (actually, it is also not marked when the url is "../ward/1/")

- Sometimes I cannot do certain assignments. In this screenshot, e.g., I get an error when moving the patient marked yellow to Room 13 and trying to save it.
image

That is currently all I've found without looking explicitly at the code. I will have a closer look at the implementation later on.

ycms/cms/templates/_raw.html
charludo: Oops :D

charludo:
> * It would be nice to have a "Cancel" button, maybe next to the "Save" button

Good idea! Done.

> * After moving multiple patients between different rooms without saving in between, I lost the overview of what I have changed in total. Maybe we can mark the edited but not yet saved assignments somehow (different background or border color?)

Also very good idea & done.

> * The sidebar entry "Ward View" is not marked as "selected" when we are at the timeline view (actually, it is also not marked when the url is "../ward/1/")

Good catch. Since this isn't just timeline-related and this PR is already pretty big, could you open a new issue? (And would you be willing to take it? I have not really looked into how the colors are set in the first place yet)

> * Sometimes I cannot do certain assignments. In this screenshot, e.g., I get an error when moving the patient marked yellow to Room 13 and trying to save it.

Oooooof. Yep, I can replicate. BUT the strange thing is: if you first assign `P. Bose` to some other room, then assign the small `B.` block to where it is shown in the picture, not only does that work, you can then place `P. Bose` back to where he originally was!! Why? How??

charludo:
> > * Sometimes I cannot do certain assignments. In this screenshot, e.g., I get an error when moving the patient marked yellow to Room 13 and trying to save it.
> 
> Oooooof. Yep, I can replicate. BUT the strange thing is: if you first assign `P. Bose` to some other room, then assign the small `B.` block to where it is shown in the picture, not only does that work, you can then place `P. Bose` back to where he originally was!! Why? How??

OK... I found a solution, but I'm not happy about it. Best I can tell, there's something weird going on in the way `exclude` and `Q` interact. IDK.

Anyways, the specific case you've shown above now works, but it would be really great if you could play around a bit and try to break it again 😂

ballerkind (approved):
Thanks for the updates. I haven't found any more issues :+1:

PR #94, opened 2023-12-06
Persistent drawers

### LINKED ISSUE #74: Remember open drawers
### Motivation

Currently, after assigning/editing a patient, we are redirected to the same view, but the sidebar is closed. It would be nicer if the user could immediately see the result of their action.

### Proposed Solution

- each time a drawer is opened, set a cookie containing the drawers unique ID
- on window load, attempt to open the drawer with that ID.
- on fail and on closing of a modal, delete the cookie (re-opening the drawer should only be attempted once)

charludo:
### Short description

Re-open drawers if they were open when the page reloaded

Just try it out, not much more to review here.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #74

### DIFF
--- a/ycms/static/src/index.ts
+++ b/ycms/static/src/index.ts
@@ -3,11 +3,11 @@ import "./css/style.scss";
 import "./js/messages";
 import "./js/patient_select_or_create";
 import "./js/utils/autocomplete";
+import "./js/utils/persistent-drawers";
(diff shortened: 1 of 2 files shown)


SnehaR26 (approved):
Works like a charm. Drawers stay open on reloading the page 🎆

Issue #70, opened 2023-11-29
Patient Detail View


charludo:
### Proposed Solution

Create a new view, link it from the patient cards in the ward view as well as from the patient list view. The new view should contain:
- all patient data as a form - no need for switching between display data / form, just use a prefilled form
- all the patients medical records, sorted descending by creation date. This should contain all information, incl. who created/updated the data when, and the record type.
- if the record is a bed assignment, also show the relevant fields
- not sure about how to handle updating the records; either also show them as a form from the beginning, or add button to enter edit mode.
- color coding the records by type would be great!

PR #100, opened 2023-12-24
Allow anonnymous patients

### LINKED ISSUE #98: Allow patients with no patient data
### Motivation

Emergency patients: don't care about data, just admit them and add data later

### Proposed Solution

Add third option in patient intake. 

Maybe also add "auto-assign bed immediately" checkbox? If the stay starts now.

charludo:
### Short description

This adds the option Mr. Mühlenbruch requested: Adding "unknown" patients. I've decided to require the approximate age and the gender of the patient for this, since these are required for a proper bed assignment and also known even when the patient is unconscious.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #98

### DIFF
--- a/ycms/cms/forms/__init__.py
+++ b/ycms/cms/forms/__init__.py
@@ -8,4 +8,5 @@ from .bed_assignment.bed_assignment_form import BedAssignmentForm
 from .intake_bed_assignment_form import IntakeBedAssignmentForm
 from .intake_record_form import IntakeRecordForm
 from .patient_form import PatientForm
+from .unknown_patient_form import UnknownPatientForm
 from .ward_form import WardForm
--- a/ycms/cms/forms/intake_bed_assignment_form.py
+++ b/ycms/cms/forms/intake_bed_assignment_form.py
@@ -47,7 +47,7 @@ class IntakeBedAssignmentForm(CustomModelForm):
         self.fields["admission_date"].initial = current_or_travelled_time()
         self.fields[
             "discharge_date"
-        ].initial = current_or_travelled_time() + datetime.timedelta(days=1)
+        ].initial = current_or_travelled_time() + datetime.timedelta(days=7)
 
     def save(self, commit=True):
         """
@@ -67,6 +67,7 @@ class IntakeBedAssignmentForm(CustomModelForm):
             creator=self.instance.creator,
             medical_record=self.instance.medical_record,
             admission_date=cleaned_data["admission_date"],
+            discharge_date=cleaned_data["discharge_date"],
             recommended_ward=cleaned_data["recommended_ward"],
             accompanied=cleaned_data["accompanied"],
         )
--- a/ycms/cms/forms/patient_form.py
+++ b/ycms/cms/forms/patient_form.py
@@ -27,4 +27,8 @@ class PatientForm(CustomModelForm):
             "date_of_birth",
             "insurance_type",
         ]
-        widgets = {"date_of_birth": forms.NumberInput(attrs={"type": "date"})}
+        widgets = {
+            "date_of_birth": forms.NumberInput(attrs={"type": "date"}),
+            "gender": forms.RadioSelect(),
+            "insurance_type": forms.RadioSelect(),
+        }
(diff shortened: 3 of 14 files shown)


SnehaR26:
Hi, it works as intended. I had a few suggestions:
- instead of approx age being a slider, can we make it in a range. 10-15, 15-20 or any custom value.
- It automatically assigns a discharge date in 7 weeks, do we need to have that field in the patient card?
- Not directly related to this but first we add the Unbekannt Unbekannt using Patient intake, then we go into ward view and manually assign the patient a bed. A feedback that Frau Büsing gave was that she expected the Manage Patients to perform the bed assignment/ discharge since the name says "manage patients". I know we do all this at a ward level, so should we think of some other name here to avoid this confusion?

charludo:
> - instead of approx age being a slider, can we make it in a range. 10-15, 15-20 or any custom value.

Yeah, I though about this too, but was unsure how to set the ranges. If we do 5 years per category, that's 20 categories for 0-100, if we do fewer, it becomes trickier to decide what to choose. Could a slider that only does increments of 5 work?

> - It automatically assigns a discharge date in 7 weeks, do we need to have that field in the patient card?

We do need the discharge date, yes. The 7 days are basically random. We can put anything else, I am open to suggestions. Realistlcally, we would want the date to be set based on the diagnosis - but I have not been able to find data for that online. If we do, that would be great though!

> - Not directly related to this but first we add the Unbekannt Unbekannt using Patient intake, then we go into ward view and manually assign the patient a bed. A feedback that Frau Büsing gave was that she expected the Manage Patients to perform the bed assignment/ discharge since the name says "manage patients". I know we do all this at a ward level, so should we think of some other name here to avoid this confusion?

Yep, I'll add options related to that in #70. That should solve a lot of the issues we faced in the roleplay (for example cancelling patients that have not yet arrived at the hospital)

charludo:
> * Not directly related to this but first we add the Unbekannt Unbekannt using Patient intake, then we go into ward view and manually assign the patient a bed. A feedback that Frau Büsing gave was that she expected the Manage Patients to perform the bed assignment/ discharge since the name says "manage patients". I know we do all this at a ward level, so should we think of some other name here to avoid this confusion?

Alright, this is partially fixed with #103, I think, but you are right, we could also rename the menu item to make it clearer - How about "Manage patient data" or even just "patient data"?

charludo:
> > - It automatically assigns a discharge date in 7 weeks, do we need to have that field in the patient card?
> 
> We do need the discharge date, yes. The 7 days are basically random. We can put anything else, I am open to suggestions. Realistlcally, we would want the date to be set based on the diagnosis - but I have not been able to find data for that online. If we do, that would be great though!

Ohh I think I misunderstood, do you mean the roleplay cards Tabea gives us?

If so, yes, there is usually a planned discharge date, if I remember correctly

charludo:
@SnehaR26 I've renamed the menu item and changed the slider step size. Would you be OK reviewing this one?

SnehaR26:
> @SnehaR26 I've renamed the menu item and changed the slider step size. Would you be OK reviewing this one?

Hi @charludo All looks good 👍 . The 'Patient Data' is an apt name and the slider is also more functional now. Just one teeny tiny thing, If as a user I have accidentally pressed the gender option in the emergency intake, I can't undo it. My only option is to reload the page (since new patient fields are disabled). If this can be corrected by a small fix then maybe we try that if not then reload is also not much of a hassle.

charludo:
> Hi @charludo All looks good 👍 . The 'Patient Data' is an apt name and the slider is also more functional now. Just one teeny tiny thing, If as a user I have accidentally pressed the gender option in the emergency intake, I can't undo it. My only option is to reload the page (since new patient fields are disabled). If this can be corrected by a small fix then maybe we try that if not then reload is also not much of a hassle.

Fair point! I've added reset buttons. 

(PS: if you go to "files changes" at the top of this PR, you can request chages, comment on specific lines, or approve from there!)

SnehaR26:
Looks perfect now!👌

Next time I will make use of the file changes to comment. Thank you 😊

SnehaR26 (approved):

PR #102, opened 2023-12-26
Add options for unassigning and ward transfers

### LINKED ISSUE #19: Request moving patients to a different station
### User Story

As a Station manager, in case my station is full I want to be able to request moving (new?) patients to a different station.

### Proposed Solution

Add functionality to request moving a patient to another station, pending acceptance of that stations manager

### Additional Context

Depends on:
- [x] #18

### LINKED ISSUE #80: Add option to unassign patient
### Motivation

This is currently not possible

### User Story

As a station manager, there are situations when I might want to unassign a patient. For example, I might be managing the intensive care ward, and when a patient is healthy enough to be transferred to a normal ward, I should be able to unassign them and move them to another ward.

### Proposed Solution

- add "unassign" button next to "cancel" button while in assignment mode
- we could also add a third button "suggest different station". Clicking this button would show a modal where you can choose the ward, then submit. In the background, the patient would be unassigned just like when clicking "unassign", and then their recommended ward would be set. Since in each ward, only no-recommendation or this-ward-recommended patients are shown, this would immediately "transfer" the patient to the new ward. This would also solve #19.

The same view can be used for both "suggest different station" and "unassign", by just passing the current station as "recommended ward" when unassigning. This also has the benefit that the patient won't be shown to managers of other stations just because we temporarily unassigned them.

### Additional Context

@noworneverev do you want to take this issue (100% OK if not, just asking because the ward view has been your thing so far 😄)

charludo:
### Short description

If you click the "Transfer" or "Assign" buttons, you will now see some additional options in the top bar, namely completely unassigning the patient (i.e. they are not discharged, but don't occupy a bed anymore) and a dropdown for moving the patient to a different ward (this will also unassign them, but also set the "recommended ward", so the patient won't be shown in the current ward's unassigned list any longer).

### Proposed changes

- implement functionality as described above
- removed a bit of javascript that was handling assignment, this is instead done by simple HTML forms now, since I felt that this was a bit more robust

### Side effects

- 
- 

### Resolved issues

Fixes: #19, #80

### DIFF
--- a/ycms/cms/templates/bed_assignment/assign_bed_assignment.html
+++ b/ycms/cms/templates/bed_assignment/assign_bed_assignment.html
@@ -16,13 +16,40 @@
                                 {% translate "Assigning patient" %} {{ patient.first_name }}
                                 {{ patient.last_name }}
                             </h2>
-                            <a href="{% url 'cms:protected:ward_detail' pk=ward.id %}">
-                                <button type="button"
-                                        class="text-white bg-red-700 hover:bg-red-800 focus:ring-4 focus:outline-none focus:ring-red-300 font-medium rounded-lg text-sm px-4 py-2 text-center inline-flex items-center me-2 dark:bg-red-600 dark:hover:bg-red-700 dark:focus:ring-red-800">
-                                    <i icon-name="x" class="mr-2"></i>
-                                    {% translate "Cancel" %}
-                                </button>
-                            </a>
+                            <div class="flex gap-2 items-center">
+                                <span class="font-bold dark:text-gray-300">{% translate "More reassignment options" %}:</span>
+                                <form method="post"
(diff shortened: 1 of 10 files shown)


ballerkind (approved):
I had a look at the code and tried different scenarios; all in all, I have nothing to complain about 👍 

The only improvement I could think of is to remove the "Unassign" button if the patient is already unassigned. But it is also fine the way it is currently.

charludo:
> The only improvement I could think of is to remove the "Unassign" button if the patient is already unassigned. But it is also fine the way it is currently.

Oh, right! Good catch, thanks. Hiding the button in that case what just one `if` block in the template, so I'm going to take the liberty of merging without a re-review 😅

PR #103, opened 2023-12-28
Add patient detail view

### LINKED ISSUE #70: Patient Detail View

charludo:
### Short description

This adds a patient detail view where we can
- edit patient data
- edit medical record + bed assignment of the current stay
- edit medical records + bed assignments of all **future** stays (important for the role play)
- create new medical records (i.e. Doctors/Nurses can add notes, lab results,...)
- see all medical history (all medical records incl. bed assignments if necessary) on a timeline

Additionally, the intake form post-save behavior has been modified:
- if a recommended ward was specified and the admission date is now or in the past, redirect to the ward and open the unassigned drawer, where the new patient is the first entry
- else, redirect to the patient detail view of the newly created patient
- reassigning/unassigning patients opens the correct, new drawer afterwards!

If you are reviewing this, please make sure to test all functionality - i.e. you will need to test it with a patient with both a current and a future stay (probbaly need to create one)

### Proposed changes

- addd the view
- reuse as much of the existing update views and forms as possible, but some new ones were necessary

### Side effects

- this PR is a bit intertwined with #102 (maybe it would be good to review/merge that one first). This can be seen in two places:
    - #102 moves the action buttons in the ward side panels around a bit and adds a link to the detail view; this link is not shown in this PR yet, you have to go through the patient list in the patient managemen view
    - #102 also changes the way discharges are handled server-side. This PR makes use of that. Right now, if you click "discharge now" in the patient detail view, you will see a json response and have to go back in the browser, then refreh the page to see the change; once #102 is merged, this will automatically happen properly, like with all other forms/actions

### Resolved issues

Fixes: #70, partially: #119


ycms/static/src/js/utils/loader.ts
```diff
@@ -9,3 +9,10 @@ window.addEventListener("beforeunload", () => {
     const TIMEOUT = 200;
     setTimeout(showLoader, TIMEOUT);
 });
+
+window.addEventListener("unload", () => {
+    const loader = document.getElementById("loader-container");
+    if (loader) {
+        loader.classList.add("hidden");
+    }
+});
```
charludo, line 18: @noworneverev We needed something like this after all 😄 Otherwise, when you go back in the browser, and the previous page had the loader, it would not disappear.

I noticed while working on this branch, so I included it here 😅

ballerkind (commented):
This works pretty well!

Some small thoughts I would like to share regarding the patient detailt view:
- When using the "Plan new stay" button and getting redirected, maybe we can pre-select the according patient in the "Search for existing patient" input
- I couldn't figure out what the purpose of the "patient intake form" record type is, especially when the result is "No hospital stay data is associated with this intake". Could you give me a use-case example? 😄 
- When searching for a diagnosis code, the dropdown menu is not shown below the input but jumps to the right.

Apart from that, it looks very nice 👍

charludo:
Thanks for the review!

> * When using the "Plan new stay" button and getting redirected, maybe we can pre-select the according patient in the "Search for existing patient" input

Yeah, fair. I added the feature. TBH it is a little hacky (and for some reason I could not get the URL parameter to register correctly in TS if I only passed the patient? Very strange), but it should be working.

> * I couldn't figure out what the purpose of the "patient intake form" record type is, especially when the result is "No hospital stay data is associated with this intake". Could you give me a use-case example? 😄

"patient intake form" is the type of record we use to signify a hospital stay. That is largely due to our DB relations:
Patient <- MedicalRecord <- BedAssignment
 
The reason I added the "...no data associated, probably cancelled..." message is that you could plan a hospital stay, but then the patient calls and cancels it (that's one of the roleplaying cards as well). In that case, an intake record exists, but the patient has never been assigned to a bed, of course.

An alternative would be to just hide those records, but tbh I kinda like that it's still included, and I imagine in a real world scenario you would also want to keep a record of the patient planning a hospital stay, but it never coming to that (for whatever reason).

But you of course have a point about showing it in this specific form. I removed the option here, and set the default to show "note"

> * When searching for a diagnosis code, the dropdown menu is not shown below the input but jumps to the right.

Interesting, this only seems to happen on Chrome. I changed the offending form to a flex layout, looks like it's fixed now.

Hey, quick question, are the fonts super bold on Chrome as well, or is this just me?? On Firefox it looks as I would expect. The font weight is set to normal/400 by default, but Chrome renders it way thicker. This only happens for that font weight, so I am unsure if maybe it's just something on my machine.

ballerkind (commented):
Thanks for updating, it works perfectly.

Regarding the font, I am not sure how it should look like but this is the way I see the page:
localhost_8086_patients_217_

One small issue I found:
If you check the screenshot, you see the text "14.01.2024 13:50, Administrator Root Root". However, I took the screenshot at 14:50. So I guess we are using the wrong timezone? Same holds for all the other times in this view

charludo:
> Regarding the font, I am not sure how it should look like but this is the way I see the page: 

OK that looks exactly like it does for me in FF. Guess my Chrome(ium) is broken. Oh well.

> One small issue I found: If you check the screenshot, you see the text "14.01.2024 13:50, Administrator Root Root". However, I took the screenshot at 14:50. So I guess we are using the wrong timezone? Same holds for all the other times in this view

Very good catch. Completely unrelated to this PR, so it is probably happening everywhere. Solution: Set the right timezone for Germany in the settings file 🤦🏼‍♀️

ballerkind (approved):
:+1:

PR #109, opened 2024-01-04
Fix sidebar color usage

charludo:
### Short description

Super tiny PR: I just noticed that we were using "absolute" color values for the sidebar-blue. Changed it the closes tailwind blue.

As a side note: Should we change the primary color to the same blue? I think this only affects some buttons, but I am starting to think that the yellow is a bit distracting (esp. in dark mode), and it could be one color less...

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #

### DIFF
--- a/ycms/static/src/css/style.scss
+++ b/ycms/static/src/css/style.scss
@@ -328,7 +328,7 @@ input[type="submit"] {
     transition: all 0.3s ease;
 
     &:hover {
-        color: #4a90e2;
+        @apply text-blue-500;
     }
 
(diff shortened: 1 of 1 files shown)


noworneverev (approved):
Agree, there's a little bit inconsistent with buttons, I think you can change if you want. I will just approve as the PR will not affect any functionality.

charludo:
> Agree, there's a little bit inconsistent with buttons, I think you can change if you want. I will just approve as the PR will not affect any functionality.

Alright, I'll open a separate PR.

PR #110, opened 2024-01-07
Get rid of Yellow

charludo:
### Short description

Replaces the yellow buttons with the blue ones we have been using everywhere. One less color should mean a bit cleaner design.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: n/a


noworneverev (approved):
🎉

PR #114, opened 2024-01-09
Add user settings

### LINKED ISSUE #107: Add user settings
### Motivation

Unnecessary

### Proposed Solution

Remove the link

### Alternatives

Just add a default user form

ballerkind:
### Short description

I decided to add a small settings view

### Proposed changes

- add settings view
- allows to change password and things like name, ward, etc.

### Side effects

- 
- 

### Resolved issues

Fixes: #107

### DIFF
--- a/ycms/cms/forms/user_form.py
+++ b/ycms/cms/forms/user_form.py
@@ -1,0 +1,17 @@
+from ..models import User
+from .custom_model_form import CustomModelForm
+
+
+class UserForm(CustomModelForm):
+    """
+    Form for editing users
+    """
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = User
+        fields = ["first_name", "last_name", "assigned_ward", "ward_as_timeline"]
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -66,8 +66,9 @@ limitations under the License.
                     <span class="block mt-5 px-5 text-gray-500 tracking-wide dark:text-gray-300">{% translate "Ward" %}</span>
                     <li>
                         {% url 'cms:protected:ward_detail_default' as url %}
+                        {% url 'cms:protected:ward_management' as ward_management_url %}
                         <a href="{{ url }}"
-                           class="{% if request.path == url %}selected{% endif %}">
+                           class="{% if request.path != ward_management_url %}{% if '/ward/' in request.path or '/timeline/' in request.path %}selected{% endif %}{% endif %}">
                             <i icon-name="bed"></i> {% translate "Ward View" %}
                         </a>
                     </li>
--- a/ycms/cms/templates/user_settings.html
+++ b/ycms/cms/templates/user_settings.html
@@ -1,0 +1,34 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load render_bundle from webpack_loader %}
+{% block title %}
+    {% translate "Settings" %}
+{% endblock title %}
+{% block content %}
+    <div class="flex flex-wrap justify-center items-start gap-8 min-h-full">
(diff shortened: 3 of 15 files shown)


ycms/cms/forms/user_form.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = User
+        fields = ["personnel_id", "email", "first_name", "last_name", "assigned_ward"]
```
charludo: IMO personnal_id should definitely not be editable. I'm ambivalent, leaning to "no" for the email as well (real-world justification: should always be tied to the email the hospital provides its employees, so no reason to change it; technical justification: in all other cases of email usage, we send confirmation links. So to stay consistent, we should send activation links to both the old and the new email, and once both links have been clicked, perform the changes. This is obviously way too much work.)

Also IDK about changing the name 😅 

Depending on how many fields remain, it could also look nicer to place the two forms below each other. If only the assigned_ward remains, I guess we could also add the view mode setting from the timeline PR eventuelly? Would make thematic sense IMO

ycms/cms/templates/_base.html
```diff
@@ -66,8 +66,9 @@
                     <span class="block mt-5 px-5 text-gray-500 tracking-wide dark:text-gray-300">{% translate "Ward" %}</span>
                     <li>
                         {% url 'cms:protected:ward_detail_default' as url %}
+                        {% url 'cms:protected:ward_management' as ward_management_url %}
                         <a href="{{ url }}"
-                           class="{% if request.path == url %}selected{% endif %}">
+                           class="{% if request.path != ward_management_url %}{% if '/ward/' in request.path or '/timeline/' in request.path %}selected{% endif %}{% endif %}">
```
charludo, line 71: Cool, thanks for fixing this!  🎉

ycms/cms/templates/user_settings.html
```diff
@@ -0,0 +1,34 @@
...
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load render_bundle from webpack_loader %}
+{% block title %}
+    {% translate "Settings" %}
+{% endblock title %}
+{% block content %}
+    <div class="flex justify-evenly items-center">
+        <div class="w-1/3 bg-white shadow p-8 dark:bg-gray-800">
```
charludo:
```suggestion
        <div class="w-1/3 rounded-lg border border-gray-300 bg-white p-4 shadow-md flex flex-col h-full dark:bg-gray-800 dark:border-gray-700">
```

I think this is copied from the user creation form view, right? We've since pivoted to using these kind of boxes everywhere. Could you also change this for the other form, and the user creation and login forms? 😇 😅

ycms/cms/views/user_settings_view.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        This function updates the settings
+        """
+        form_type = request.POST["form_type"]
+
+        if form_type == "general":
+            user_form = UserForm(data=request.POST, instance=request.user)
+            if user_form.is_valid():
+                user_form.save()
+                messages.success(request, _("Your settings were successfully updated!"))
```
charludo, line 38: I'm generally not a big fan of this pattern tbh. Any hardcoded string is a potential future error source.

Keeping two separate forms makes sense in this case, since the `PasswordChangeForm` does do some heavy lifting, but how about adding a separate view for it to `post` to? All other password-related views live in the `authentication` subfolder.

charludo (changes requested):
Very cool, thank you! :tada: I just have a couple of small comments.

charludo (approved):
Awesome, this is good to go.

Thanks for finding and fixing the spelling mistakes - those have gone unoticed for a WHILE, haha

charludo (changes requested):
Argh, sorry, just noticed: Please either:
- run `prune_database.sh`, then `run.sh`. This will create a new migration. It needs a docstring (and maybe a better file name)
- or: fix the spelling in the original migration... Not super great practice, but for a spelling fix....

charludo (approved):

PR #118, opened 2024-01-11
Fix overflowing progress bars

charludo:
### Short description

The overflow was occurring when the release day is today, but "not yet", i.e. at some time later today.

### Proposed changes

- fix progress overflow
- also tweak how accompanying persons are displayed in the room card, by always fixing them to the right column. This should either show them next to the person in the room as before, or, if through algorithmic assignments 3 people are in a 2 bed room, it shows the accompanying person under the correct patient

### Side effects

- 
- 

### Resolved issues

Fixes: n/a


noworneverev (approved):
I've tested it out, and it looks good.⭐

PR #123, opened 2024-01-13
add collapsible sidebar

### LINKED ISSUE #122: make sidebar collapsible in large screen
### Motivation

The sidebar cannot be collapsed on large screens, which makes it difficult to see all rooms clearly.

noworneverev:
### Short description

Make sidebar collapsible in large screens.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #122

### DIFF
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -24,7 +24,7 @@ limitations under the License.
             <i id="closeIcon" class="hidden" icon-name="x"></i>
         </button>
         <div id="sidebar"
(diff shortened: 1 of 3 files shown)


charludo (approved):
Very nice! I especially like that you are saving to `localStorage`, really helps with a consistent feel, nice.

~One nitpick: I was initially confused where the toggle button is. The single `/` in the middle of the screen height is very easy to miss.~

~Could you maybe just re-use the mobile toggle button, also putting it at the top of the screen?~

~I guess it would be a fair bit more prominent though, hm. If you want to keep it as low profile as it is, how about using something like this: https://lucide.dev/icons/arrow-right-from-line in place of the `/`?~

Actually, nevermind. The more I look at it, the more I like it.

PR #126, opened 2024-01-13
Add room & bed creation to the ward form

### LINKED ISSUE #77: Add the functionality to create and add rooms and beds for a ward
### Motivation

Currently ZBM user can only add a ward . We want them to be able to add rooms and subsequently beds in those rooms.

### User Story

For achieving a complete functionality of our product, we should have this feature.
### Proposed Solution

Adding a form / appending ward form
### Additional Context

should be able to add multiple beds/rooms with a single submission. User shouldn't have to click multiple times.

SnehaR26:
### Short description

Adds options for adding rooms and beds to a new ward.

### Proposed changes

-  add a button by which new rooma can be added. Rooms have a default name and two default beds 
- rooms and beds can be added, removed, and changes at will.
- this PR does **NOT** add support for editing existing wards. It could be done with some work in a new PR, almost entirely re-using the code from here, but I do not think the work is justified before the deadline.

### Side effects

- none AFAIK

### Resolved issues

Fixes: #77


SnehaR26:
On an unrelated note , I am unable to push through git so I added the changes via github

image

charludo:
> On an unrelated note , I am unable to push through git so I added the changes via github
> 
> image

The error message gives you a pretty good hint, I think: authentication for things like pushing via `https` have been removed from github (for security reasons). Instead, you should be using `ssh` urls, i.e. instead of
`https://github.com/charludo/ycms.git` you should be using
`git@github.com:charludo/ycms`

You can probably switch by doing `git remote set-url origin git@github.com:charludo/ycms`

However, please be aware that you need to setup SSH authentication first. Githubs documentation has you covered: https://docs.github.com/en/authentication/connecting-to-github-with-ssh

ycms/cms/forms/bed_addition_form.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = Bed
+        fields = ["number_beds"]
```
charludo: The `Bed` model does not have a field `number_beds` (and neither do `Room` or `Ward`). This should probably be an extra field on the existing `WardForm`. Take a look at the `UnknownPatientForm` - there you can see an example of how an additional field which does not exist on the model can be added to the form (`approximate_age`, in this case).

ycms/cms/templates/ward/ward_management.html
```diff
@@ -62,6 +62,11 @@
                         {% endfor %}
                         <button type="submit" class="btn mt-4">{% translate "Add New Ward" %}</button>
                     </form>
+                    <ul>
+                    <li><label for="nrroom">Number of beds You want to add</label></li>
+                <li><input type="number" id="nrroom"/></li>
+                    </ul>
+
```
charludo: If you add the field via the correct form, this can be removed, since just above this code, we are iterating over and displaying all form fields.

charludo:
So from what I am seeing, what you are trying to do is:
- add a number field `number_beds` on the `WardForm`
- upon form submission, create `number_beds`/2 new rooms
- for each of those newly created rooms, create two beds

right?

Some things to consider:
- each bed needs a `bed_type` - you could just set them to `bed_types.NORMAL`, of course, even though choice would be nice
- each room needs a `room_number` - how are you planning on generating them?
- not being able to create rooms with 1, 3, n beds is kind of sad, since the wards do support them

If you wanted to give the user more choice, while also avoiding the name generation issue:
- show an "add room" button in the form
- use javascript to listen for clicks on that button (lots of examples already in the code)
- whenever the button is clicked, use javascript to insert a small form consisting of a `room_number` and `number_beds` input
- when the `WardForm` is submitted, use JS to prevent the submission, iterate over all the room forms, and create a JSON object looking something like this: `{ "room_number-1": 2, "another-room-number": 1, "04": 3}`, of course depending on what the user set as values
- then submit the form using JS, including your newly built array (for example in a hidden input field)
- in the `post` method of the `WardManagementView`, you receive this value, and can now just iterate over the list entries, creating a room with the given name, plus the given amount of beds for the room.

This still wouldn't allow setting the bed type, although it could pretty easily be extended to include that functionality.

SnehaR26:
> The error message gives you a pretty good hint, I think: authentication for things like pushing via `https` have been removed from github (for security reasons). Instead, you should be using `ssh` urls, i.e. instead of `https://github.com/charludo/ycms.git` you should be using `git@github.com:charludo/ycms`
> 
> You can probably switch by doing `git remote set-url origin git@github.com:charludo/ycms`
> 
> However, please be aware that you need to setup SSH authentication first. Githubs documentation has you covered: https://docs.github.com/en/authentication/connecting-to-github-with-ssh

Thank you, I was able to set up a new key and it seems to work for now. 
For the room addition, yup I must consider all this. Let me try with your comments

ballerkind (approved):
Works as intended :+1:

PR #120, opened 2024-01-11
Cleanup of the test data loading

charludo:
### Short description

In anticipation of the final presentation data, I've done some cleanup of the way we import new and old test data. Feel free to review, of course, but the management command is not relevant to the normal operation of the project.

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #

PR #127, opened 2024-01-13
Feature/responsive design

### LINKED ISSUE #125: Responsive design
### Motivation

If we want users to be able to try this on their own devices, the design must be responsive. We are on a pretty good track to achieving this, but some views need some attention.

### Additional Context

The Timeline view will probably never be usable on mobile. We can try to make everything fit, but it is not gonna look good.

noworneverev:
### Short description

Improve the responsiveness of the website

### Proposed changes

- Fix the sidebar and add a backdrop when opening it on a small screen
- Adjust buttons on patient cards for small screens
- Adjust the font size in patient data (font size was too large in small screens)
- Fix user creation in small screens
- Fix time travel control in small screens
- Fix login form in small screens
- Fix header in page of bed assignment
- Fix ward selection header
- Fix success/error notifications' position in small screens
- Fix timeline view header in small screens
- Adjust patient list table
- Fix password reset request and password request confirm templates in small screens
- Fix responsiveness of ward management

~Pending:~
~- overlap when opening sidebar in patient intake, already know that if removing all position: relative can solve the issue, including tom-select's position: relative (don't know how to remove this one)~

### Side effects

- 
- 

### Resolved issues

Fixes: #125

### DIFF
--- a/ycms/cms/templates/_raw.html
+++ b/ycms/cms/templates/_raw.html
@@ -43,7 +43,7 @@ limitations under the License.
     <body class="font-default text-lg bg-background-50">
         {% if messages and not messages_suppressed %}
             <div id="messages"
-                 class="absolute top-32 right-8 flex flex-col max-w-screen-sm z-50">
+                 class="absolute top-8 right-8 flex flex-col max-w-screen-sm z-50">
                 {% for msg in messages %}
                     {% if msg.level_tag == 'info' %}
                         <div class="message bg-blue-100 border-l-4 border-blue-500 dark:bg-blue-700 dark:border-blue-400"
--- a/ycms/cms/templates/timeline/timeline.html
+++ b/ycms/cms/templates/timeline/timeline.html
@@ -7,34 +7,37 @@
 {% block content %}
     <div class="flex justify-center items-center">
         <div class="container mx-auto">
-            <div class="col-span-full rounded-lg border border-gray-300 bg-white shadow-md flex flex-nowrap h-full items-center justify-between p-4 mb-4 hover:bg-gray-100 dark:bg-gray-800 dark:border-gray-700 dark:hover:bg-gray-700">
+            <div class="flex flex-col gap-2 xl:flex-row rounded-lg border border-gray-300 bg-white shadow-md flex flex-nowrap h-full items-center justify-between p-4 mb-4 hover:bg-gray-100 dark:bg-gray-800 dark:border-gray-700 dark:hover:bg-gray-700">
                 {% include "ward/ward_selection_form.html" %}
-                <div class="flex gap-2">
-                    <div id="change-counter"
-                         class="hidden flex items-center gap-2 dark:text-white">
-                        <span>0</span> {% translate "unsaved bed assignment changes" %}
+                <div id="change-counter"
+                     class="hidden flex flex-col lg:flex-row items-center gap-2 dark:text-white">
+                    <div class="shrink-0">
+                        <span>0</span>
+                        {% translate "unsaved bed assignment changes" %}
+                    </div>
+                    <div class="grid grid-cols-2 gap-2 w-full">
                         <form enctype="multipart/form-data"
                               method="post"
                               id="timeline-changes-form"
-                              class="flex items-center">
+                              class="flex items-center w-full">
                             {% csrf_token %}
                             <input type="hidden" name="timeline_changes" id="timeline-changes" />
-                            <button class="text-white bg-green-500 hover:bg-green-700 focus:ring-4 focus:outline-none focus:ring-green-300 rounded inline-flex items-center px-5 py-2.5 text-center">
+                            <button class="w-full text-white bg-green-500 hover:bg-green-700 focus:ring-4 focus:outline-none focus:ring-green-300 rounded inline-flex items-center justify-center px-5 py-2.5 text-center">
                                 {% translate "Save" %}
                             </button>
                         </form>
                         <a href="{% url "cms:protected:timeline" selected_ward_id %}"
-                           class="text-red-700 border-red-700 border-2 hover:border-red-800 dark:text-white hover:text-white hover:bg-red-800 focus:ring-4 focus:outline-none focus:ring-red-300 rounded px-5 py-2 dark:hover:bg-red-700 dark:focus:ring-red-800">
+                           class="w-full lg:w-auto lg:col-span-1 text-center text-red-700 border-red-700 border-2 hover:border-red-800 dark:text-white hover:text-white hover:bg-red-800 focus:ring-4 focus:outline-none focus:ring-red-300 rounded px-5 py-2 dark:hover:bg-red-700 dark:focus:ring-red-800">
                             {% translate "Cancel" %}
                         </a>
+                        {% if "/suggest/" not in request.path %}
+                            <a href="{% url "cms:protected:timeline_suggest" selected_ward_id %}"
+                               class="btn w-full col-span-2">
+                                <i icon-name="bot" class="mr-2 align-text-bottom"></i>
+                                {% translate "Suggest Assignment" %}
+                            </a>
+                        {% endif %}
                     </div>
-                    {% if "/suggest/" not in request.path %}
-                        <a href="{% url "cms:protected:timeline_suggest" selected_ward_id %}"
-                           class="btn">
-                            <i icon-name="bot" class="mr-2 align-text-bottom"></i>
-                            {% translate "Suggest Assignment" %}
-                        </a>
-                    {% endif %}
                 </div>
                 {% with checked=True %}
                     {% include "ward/ward_mode_switch.html" %}
--- a/ycms/cms/templates/timetravel_controls.html
+++ b/ycms/cms/templates/timetravel_controls.html
@@ -1,25 +1,31 @@
 {% load i18n %}
 <div id="timetravel-controls"
      class="hidden fixed bottom-2 left-1/2 transform -translate-x-1/2 flex justify-center items-stretch bg-background-50 border border-gray-500 rounded-xl shadow-lg z-50 p-4 text-md gap-2 dark:bg-background-900">
-    <button class="btn btn-outline-plain" data-minutes="-1440">
+    <button class="hidden md:!inline-block btn btn-outline-plain"
+            data-minutes="-1440">
         <i icon-name="chevrons-left"></i>{% translate "day" %}
     </button>
-    <button class="btn btn-outline-plain" data-minutes="-60">
+    <button class="hidden md:!inline-block btn btn-outline-plain"
+            data-minutes="-60">
         <i icon-name="chevrons-left"></i>{% translate "hour" %}
     </button>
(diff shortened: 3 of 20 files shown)


charludo:
Awesome, thank you so much for tackling this issue.

Really, really nice progress already. I noticed a couple of additional things:
- ward view: the ward selection header is overflowing. Maybe the timeline toggle can just be placed an a second line?
- ward view sidebar: clicking "discharge" seems to bring up the confirmation modal (background overlay is shown), but somewhere out of frame
- the success/error notifications cling to the left side of the screen
- timeline view: same thing as ward view header, plus when a change was made, it overflows even further
- the timeline seems to have a higher z-index than the mobile menu (it's visible over the menu)
- ward management: the form should probably just not be shown as flex anymore on mobile, same with the "at a glance" section; not sure what to do with the table though
- same thing with the patient list table. I just tried it out, you *could* give each `tr` and `td` element a class like `block` on small screens, then the table structure is at least broken
- the same changes you made to the login form probably should also be made to the password reset request and password request confirm templates

> overlap when opening sidebar in patient intake, already know that if removing all position: relative can solve the issue, including tom-select's position: relative (don't know how to remove this one)

As far as I can tell, it's enough to set a higher z-index on both the `#sidebar` and `#sidebarBackdrop`. They are currently set to `z-100`, try `z-[999]` and `z-[998]`, respectively. This also fixes the issue for the timeline!

noworneverev:
@charludo Not sure why the test fails after rebasing the latest develop branch. It can be solved by adding checks to cached property. Should I modify this?

```python
    @cached_property
    def current_admission_date(self):
        """
        Helper property for accessing the patient's current admission date

        :return: the current admission date
        :rtype: str
        """
        if not self.current_stay:
            return None
        return self._get_date_status(self.current_stay.admission_date)

    @cached_property
    def current_discharge_date(self):
        """
        Helper property for accessing the patient's current discharge date

        :return: the current discharge date
        :rtype: str
        """
        if not self.current_stay:
            return None
        return self._get_date_status(self.current_stay.discharge_date)
```

charludo:
> @charludo Not sure why the test fails after rebasing the latest develop branch. It can be solved by adding checks to cached property. Should I modify this?
> 
> ```python
>     @cached_property
>     def current_admission_date(self):
>         """
>         Helper property for accessing the patient's current admission date
> 
>         :return: the current admission date
>         :rtype: str
>         """
>         if not self.current_stay:
>             return None
>         return self._get_date_status(self.current_stay.admission_date)
> 
>     @cached_property
>     def current_discharge_date(self):
>         """
>         Helper property for accessing the patient's current discharge date
> 
>         :return: the current discharge date
>         :rtype: str
>         """
>         if not self.current_stay:
>             return None
>         return self._get_date_status(self.current_stay.discharge_date)
> ```

Hey, sorry for not seeing this earlier. Apologies, this is my fault. I forgot to add a test patient which we could use for tests indefinitely. One of the patients That were used in the tests must have been discharged yesterday evening.

I have [added this test case here](https://github.com/charludo/ycms/pull/120/commits/e0baac5c2114b3ec8ddad9c02ac3bd8a050a706f). It's a single patient which will stay in the hospital for another 5 years... 😄 That should hopefully do it ^^

The first thing `_get_date_status()` does is check if the passed date is `None`, so in actual usecases we are already safe. IDK if the additional checks are necessary, since this time, they did uncover an actual error. Then again, they don't hurt either, so... Your choice if you want to keep them in.

charludo (changes requested):
Nice! Again, thank you so much for doing this issue.

I have a couple small requests:

- the ward/timeline switch is much better now, except it's not aligned with the "Ward" title above it. That one has a x-padding of 4, maybe we could remove that on small screens?
- when assigning/transfering a patient, could you break up the "More reassignment options" and the drop down? Just setting `md:flex` or `lg:flex` instead of just `flex` does the trick. And then the "Unassign" button below it should also lose its left padding on small screens
- the success/error notifications still cling to the left of the screen. Setting the top and right values to `md:top-8 md:right-8` on `#messages` in `_raw.html` does the trick.
- the number of unsaved bed assignments in the ward view is still unreadable on mobile. Setting `flex-wrap` on `#change-counter` and wrapping the span and the "unsaved bed assignment changes" in a div works, I think. Alternatively, if you find a way to also make the 2 buttons take the full width together, that would look ieven nicer, but not necessary
- the header of the patient list can probably just be hidden on mobile
- I get what you did with the table, but it's really necessary that we are still able to access the details from there 😅 I get that this table is a fair bit of work though, if you think it's out of scope for this PR, feel free to instead create a new issue.

ycms/cms/templates/timetravel_controls.html
```diff
@@ -1,25 +1,27 @@
...
+        <button class="btn btn-outline-plain " data-minutes="1">
+            {% translate "minute" %}<i icon-name="chevrons-right"></i>
+        </button>
+        <button class="btn btn-outline-plain " data-minutes="60">
+            {% translate "hour" %}<i icon-name="chevrons-right"></i>
+        </button>
+        <button class="btn btn-outline-plain " data-minutes="1440">
+            {% translate "day" %}<i icon-name="chevrons-right"></i>
+        </button>
+    </div>
```
charludo:
```suggestion
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="-1440">
        <i icon-name="chevrons-left"></i>{% translate "day" %}
    </button>
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="-60">
        <i icon-name="chevrons-left"></i>{% translate "hour" %}
    </button>
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="-1">
        <i icon-name="chevrons-left"></i>{% translate "minute" %}
    </button>
    {# djlint:off #}
    <input type="datetime-local" style="width: auto !important; box-shadow: none !important;" />
    {# djlint:on #}
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="1">
        {% translate "minute" %}<i icon-name="chevrons-right"></i>
    </button>
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="60">
        {% translate "hour" %}<i icon-name="chevrons-right"></i>
    </button>
    <button class="hidden md:!inline-block btn btn-outline-plain"
            data-minutes="1440">
        {% translate "day" %}<i icon-name="chevrons-right"></i>
    </button>
```

What do you think about this? The controls take up quite a lot of screen space on mobile, and since the datepicker is technically enough to set everything and the buttons are just for convenience, I thought this might work, too?

noworneverev: Looks good, but I just noticed that when I choose a specific time, after clicking outside to apply the change, the time displayed on the control decreases by one hour. This is the bug right?

charludo: That might be related to the change in timezone in #103 

I'm currently working on a PR to fix a bunch of bugs with the time travelling, guess I'll add it to the list 😅

Thanks for noticing and pointing it out!

noworneverev:
> Nice! Again, thank you so much for doing this issue.
> 
> I have a couple small requests:
> 
> * the ward/timeline switch is much better now, except it's not aligned with the "Ward" title above it. That one has a x-padding of 4, maybe we could remove that on small screens?

Do You mean `space-x-4` in `ward_selection_form.html`? If so, I think this one is necessary, otherwise the select element is too close to the word "Ward".

> * when assigning/transfering a patient, could you break up the "More reassignment options" and the drop down? Just setting `md:flex` or `lg:flex` instead of just `flex` does the trick. And then the "Unassign" button below it should also lose its left padding on small screens

ok

> * the success/error notifications still cling to the left of the screen. Setting the top and right values to `md:top-8 md:right-8` on `#messages` in `_raw.html` does the trick.

But I already set top and right like follows, I did't see any different after I add  `md:top-8 md:right-8`.

```html
            <div id="messages"
                 class="absolute top-8 right-8 flex flex-col max-w-screen-sm z-50">
```

> * the number of unsaved bed assignments in the ward view is still unreadable on mobile. Setting `flex-wrap` on `#change-counter` and wrapping the span and the "unsaved bed assignment changes" in a div works, I think. Alternatively, if you find a way to also make the 2 buttons take the full width together, that would look ieven nicer, but not necessary

ok

> * the header of the patient list can probably just be hidden on mobile

I think it's needed, no reason to show different context in different screen size, except the layout.

> * I get what you did with the table, but it's really necessary that we are still able to access the details from there 😅 I get that this table is a fair bit of work though, if you think it's out of scope for this PR, feel free to instead create a new issue.

It's tricky to make the table responsive. This is the best I can do for now, but I just noticed that on a small screen, the fields, except for the name field, can not be edited since I intentionally hid them.

charludo:
> Do You mean `space-x-4` in `ward_selection_form.html`? If so, I think this one is necessary, otherwise the select element is too close to the word "Ward".

No, in `ward_selection_form`, replace `p-4` with `md:px-4 py-4` (right next to `space-x-4`).

Without: 
image
With: 
image

ycms/cms/templates/_raw.html
```diff
@@ -43,7 +43,7 @@
     <body class="font-default text-lg bg-background-50">
         {% if messages and not messages_suppressed %}
             <div id="messages"
-                 class="absolute top-32 right-8 flex flex-col max-w-screen-sm z-50">
+                 class="absolute top-8 right-8 flex flex-col max-w-screen-sm z-50">
```
charludo, line 46: > But I already set top and right like follows, I did't see any different after I add `md:top-8 md:right-8`.
> 
> ```
>             <div id="messages"
>                  class="absolute top-8 right-8 flex flex-col max-w-screen-sm z-50">
> ```

```suggestion
                 class="absolute md:top-8 md:right-8 flex flex-col max-w-screen-sm z-50">
```

Without: 
image

With:
image

ycms/cms/templates/timeline/timeline.html
```diff
@@ -7,25 +7,30 @@
...
-                    <form enctype="multipart/form-data"
-                          method="post"
-                          id="timeline-changes-form"
-                          class="flex items-center">
-                        {% csrf_token %}
-                        <input type="hidden" name="timeline_changes" id="timeline-changes" />
-                        <button class="text-white bg-green-500 hover:bg-green-700 focus:ring-4 focus:outline-none focus:ring-green-300 rounded inline-flex items-center px-5 py-2.5 text-center">
-                            {% translate "Save" %}
+                     class="hidden flex flex-col lg:flex-row items-center gap-2 dark:text-white">
+                    <div class="flex flex-row gap-2 w-full">
```
charludo:
```suggestion
                    <div class="shrink-0">
```
No difference on mobile, but on Desktop:

Without: 
image

With: 
image

charludo:
> It's tricky to make the table responsive. This is the best I can do for now, but I just noticed that on a small screen, the fields, except for the name field, can not be edited since I intentionally hid them.

Alright, then please open a new issue for that.

charludo:
Nice! Please squash the commits, and then it's ready to merge.

charludo (approved):

PR #129, opened 2024-01-16
Fix timetravel bugs

### LINKED ISSUE #71: Timetravelling does not work for newly created patients/bed assignments
### Additional Information

<details>
  <summary>Traceback</summary>

  ```
  ```
</details>

### LINKED ISSUE #92: Make timeline work with timetravel

charludo:
### Short description

Fixes various timetravel bugs (see below).

In short: timetravelling works, as long as you just want to SEE. If you want to CHANGE things in the past or in the future..... Yeah I think this is not worth the effort.

Please check out any and all edge cases you can think of while keeping the above in mind 😅 

### Proposed changes

- the DB is now properly refreshed after leaving timetravel; previuosly, the last timetravelled-to objects were still shown
- the time entered in the output field no longer differs by 1 hour after redirecting (thanks for pointing this out, @noworneverev)
- the time travelled to is no longer offset by one hour compared to how the objects are filtered (separate bug, lol) 

### Side effects

- 
- 

### Resolved issues

Fixes: #71, #92 (I guess...)

### DIFF
--- a/ycms/cms/models/room.py
+++ b/ycms/cms/models/room.py
@@ -53,7 +53,9 @@ class Room(AbstractBaseModel):
         accompanied_count = sum(
             1 for patient in self.patients() if patient.current_stay.accompanied
         )
-        return sum(1 for bed in self.beds.all() if bed.is_available) - accompanied_count
+        return max(
+            0, sum(1 for bed in self.beds.all() if bed.is_available) - accompanied_count
+        )
 
     @cached_property
     def occupied_beds(self):
--- a/ycms/cms/models/timetravel_manager.py
+++ b/ycms/cms/models/timetravel_manager.py
@@ -1,5 +1,6 @@
 import logging
 import threading
+from datetime import timedelta
 
 from django.core.exceptions import FieldError
 from django.db import models
@@ -26,9 +27,10 @@ class TimetravelManager(models.Manager):
         queryset = super().get_queryset()
 
         request = _thread_locals.request if hasattr(_thread_locals, "request") else None
-        if not (request and (time := parse_datetime(request.GET.get("time")))):
+        if not (request and request.GET.get("time")):
             return queryset
 
+        time = current_or_travelled_time()
         timetravel_filters = [
             "created_at__lte",
             "admission_date__lte",
@@ -48,7 +50,7 @@ def current_or_travelled_time():
     whenever we are timetravelling
     """
     try:
-        return parse_datetime(_thread_locals.request.GET["time"])
+        return parse_datetime(_thread_locals.request.GET["time"]) - timedelta(hours=1)
     except AttributeError:
         return timezone.now()
 
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -118,13 +118,14 @@ limitations under the License.
                 </ul>
                 <div class="text-center text-gray-600 w-full mt-auto">
                     <li class="mb-4" id="timetravel-start">
-                        <a href="?time=2024-01-01T00%3A00%3A00.000Z" class="justify-center">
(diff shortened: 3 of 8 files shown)


noworneverev (commented):
Looks really nice, but I have some questions:
- When I time travel in ward view, then I switch to timeline mode, the time travel is cancelled and I'm back to the current time. I think this behavior is expected right? If so maybe disable the switch button when user starts time travel. It's unexpected for me that when I switch to timeline and I suddenly back to today.
- Is it possible to focus on the date time travel bring me to in timeline mode? When I choose a date, I don't know if it succesfully bring me to the future or past, as it always focus me on today.

charludo:
>     * When I time travel in ward view, then I switch to timeline mode, the time travel is cancelled and I'm back to the current time. I think this behavior is expected right? If so maybe disable the switch button when user starts time travel. It's unexpected for me that when I switch to timeline and I suddenly back to today.

Fair point. The reason this happens is that multiple redirects happen after clicking the switch, and the GET parameters get lost. I'm sure it can be fixed, but IDK if it is worth the trouble; deactivating the switch sounds like a good idea.

>     * Is it possible to focus on the date time travel bring me to in timeline mode? When I choose a date, I don't know if it succesfully bring me to the future or past, as it always focus me on today.

Similarly here, I'm sure it's possible, but the timeline has its own way of setting the date. Do you think this is worth trying? The laternative is just not showing the timetravel link in the timeline. After all, the timeline lets us see "all of time", and there's little point in timetravelling there 😄

charludo:
Could you check if the new solution works for you?

noworneverev:
> > ```
> > * When I time travel in ward view, then I switch to timeline mode, the time travel is cancelled and I'm back to the current time. I think this behavior is expected right? If so maybe disable the switch button when user starts time travel. It's unexpected for me that when I switch to timeline and I suddenly back to today.
> > ```
> 
> Fair point. The reason this happens is that multiple redirects happen after clicking the switch, and the GET parameters get lost. I'm sure it can be fixed, but IDK if it is worth the trouble; deactivating the switch sounds like a good idea.
> 

Sorry I just found it's frustrated that I cannot click the button whenever I want, I prefer the version before this commit😅.

> > ```
> > * Is it possible to focus on the date time travel bring me to in timeline mode? When I choose a date, I don't know if it succesfully bring me to the future or past, as it always focus me on today.
> > ```
> 
> Similarly here, I'm sure it's possible, but the timeline has its own way of setting the date. Do you think this is worth trying? The laternative is just not showing the timetravel link in the timeline. After all, the timeline lets us see "all of time", and there's little point in timetravelling there 😄

I test the timeline, but it can only see the current time onward, I cannot see past stays in timeline.

charludo:
> I test the timeline, but it can only see the current time onward, I cannot see past stays in timeline.

Oh, right! Forgot about that. 

Is this important enough to spend time on? Frankly, I thought the timetravel was a bit of a gimmick, and there's no practical usecase I can think of for any of the roles we have to need to be able to check all past stays at once (individual patient stays can of course be checked in the patient details, without timetravel).

If you want to, feel free to work on this (we actually have an issue, #92), but I will not be putting more time into the timetravel 😅 
And since this is its own feature (kind of), I don't think it's in scope for this PR.

charludo:
> > I test the timeline, but it can only see the current time onward, I cannot see past stays in timeline.

> Oh, right! Forgot about that.

Just to get back on this: It's not a bug, it's to speed up the loading time. Also, we would need to implement a way to lock past bed assignments from being changed retroactively, and filter all of them out before sending them to the solver. IMO it's just a lot of work for virtually no benefit.

noworneverev (approved):
Perfect! 🎉

PR #132, opened 2024-01-16
Make patient list table responsive

### LINKED ISSUE #130: fix responsiveness of the table
### Motivation

Although #127 has already add some responsivenss for the table in `patients_list.html`, it introduced a new bug, and now in small screens users can longer edit. 

### Proposed Solution

Possible solutions:
- Use only `overflow-x-auto` for div that wrap the table, revert the modification inside table.
- Use the same appearance as the `add-paitent` section

charludo:
### Short description

Make table in the patinets list responsive.

### Proposed changes

- on mobile, show a "list of cards" rather than an actual table
- make some changes to the timeline header on desktop/mobile, partly reverting changes from #127  

### Side effects

- on screen sizes starting from `lg` to somewhere between `xl` and `xxl`, the table scrolls. I think it's acceptable... Not many devices in that size category, esp. during the presentation... 😅 

### Resolved issues

Fixes: #130

### DIFF
--- a/ycms/cms/templates/_raw.html
+++ b/ycms/cms/templates/_raw.html
@@ -43,7 +43,7 @@ limitations under the License.
     <body class="font-default text-lg bg-background-50">
         {% if messages and not messages_suppressed %}
             <div id="messages"
-                 class="absolute top-8 right-8 flex flex-col max-w-screen-sm z-50">
+                 class="absolute md:top-8 md:right-8 flex flex-col max-w-screen-sm z-50">
                 {% for msg in messages %}
                     {% if msg.level_tag == 'info' %}
                         <div class="message bg-blue-100 border-l-4 border-blue-500 dark:bg-blue-700 dark:border-blue-400"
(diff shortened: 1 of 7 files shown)


ballerkind (changes requested):
The idea with the cards instead of table rows looks very clean.

However, the edit button in the patient-data view does not toggle the gender field correctly, when using the regular desktop size. Both the gender label as well as the "radio" input are shown.

Also, the "Save" and "Cancel" button in the timeline view are above each other on a desktop screen. I think it would be nicer for them to be in one line with the "Suggest Assignment" button

charludo:
> However, the edit button in the patient-data view does not toggle the gender field correctly, when using the regular desktop size. Both the gender label as well as the "radio" input are shown.

Oof, good catch. Fixed.

> Also, the "Save" and "Cancel" button in the timeline view are above each other on a desktop screen. I think it would be nicer for them to be in one line with the "Suggest Assignment" button

Also good catch. Seems like I was a bit too focused on mobile! Should be fixed as well now.

ballerkind (approved):
:+1:

PR #135, opened 2024-01-16
fix sidebar responsiveness

### LINKED ISSUE #134: fix sidebar responsivenesss
### Motivation

After adding the document to the sidebar #133 , it's a little bit crowded for sidebar in small screens, because of the bottom fixed elements. It become even worse when minimize the window.

### Screenshots

image
image

noworneverev:
### Short description

Fix sidebar responsiveness in small screens

### Proposed changes

- 
- 

### Side effects

- 
- 

### Resolved issues

Fixes: #134

### DIFF
--- a/ycms/cms/templates/_base.html
+++ b/ycms/cms/templates/_base.html
@@ -31,167 +31,171 @@ limitations under the License.
                     <i id="closeIcon" icon-name="x"></i>
                 </button>
             </div>
-            <a href="{% url 'cms:protected:index' %}"
-               class="block border-b dark:border-gray-600 flex items-center text-gray-600 dark:text-white">
-                {% if is_dark_theme %}
-                    <img src="{% static 'images/logo-bed.svg' %}"
-                         width="80px"
-                         height="auto"
-                         alt="{% translate " Logo" %}"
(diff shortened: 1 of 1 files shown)


ycms/cms/templates/_base.html
```diff
@@ -115,83 +115,85 @@
                     </a>
                 </li>
             </ul>
-            <div class="absolute bottom-0 w-full text-center text-gray-600">
+            <div class="text-center text-gray-600 xl:absolute xl:bottom-0 xl:w-full">
```
charludo: OK, I hope I can describe it so this makes sense! 😄 

### Step 1:
```suggestion
            <div class="text-center text-gray-600 w-full mt-auto">
```

### Step 2:
Below the closing `</div>` in line 33, insert `<div class="flex flex-col justify-between min-h-full">`

### Step 3:
That div then has to close in line 174, just after the `© YCMS 2023` line (I guess it should be 2024 now?? 😄)


Then on low-height screens, nothing overlaps, and I get to keep my sticky footer 😋

noworneverev: Great, I tested it on different models of phone, all looks good.

charludo (approved):
:tada:

PR #136, opened 2024-01-16
:tada: Public Hosting :tada:

### LINKED ISSUE #28: Host an instance publicly

charludo:
# ⚠️ THIS PR WILL NOT BE MERGED ⚠️ 

Please still feel free to review it. But this PR makes the entire application quite insecure. It is only intended for the final presentation, so users will not have to deal with usernames and passwords.

Fixes: #28


ycms/cms/templates/authentication/demo_login.html
noworneverev, line 1: image

The top section is cut when viewed on my iPAD with horiziontal view.

noworneverev: I think we can use the tablet in presentation, but stick to the vertical view. I just noticed some weird behavior in horizontal view, and it's not worth time fixing.

charludo: Well, at least the demo login you screenshotted is fixed now.

But yes, I agree that it's pretty unlikely that any of the attendants will want to try on a landscape mode tablet.

noworneverev:
I think the room numbers in the other two wards (Pädiatrie and Intensivmedizin) should be corrected, or is this intentional?
image
image

charludo:
> I think the room numbers in the other two wards (Pädiatrie and Intensivmedizin) should be corrected, or is this intentional? image image

It's correct, in the sense that for Intensivmedizin, we got the room numbers from Tabea. (Also room number != room ID for us).

Think of it like this, they are rooms Pädiatrie-15 and Intensivmedizin-15 🤷🏼‍♀️

